PR #62, opened 2016-06-17
54 update colours

fredkingham:


fredkingham:
@davidmiller I'm merging this because it makes things look nice and I wanted Gabriel's perspective on the table display

PR #63, opened 2016-06-17
52 blood cultures form

fredkingham:


fredkingham:
this is deployed http://elcid-rfh-test.openhealthcare.org.uk/#/patient/2/2

still got 4 bugs
- [x] it looks like antimicrobials aren't saving (or aren't displaying, but probably the former)
- [x] the demographics modal isn't working (pending a merge of a pull request in to opal v0.6.0 then merge up to 7, also a merge up of referrals if necessary)
- [x] still got a faint green line underneath the main header, looks like a border property needs to be removed/updated somewhere
- [ ] date positive as of Gabriel's diagram should be on the isolate (actually it should be on the bottom, but for the time being, especially give how we're displaying it, bottle makes more sense)

2 issues
- [ ] microscopy is using a foreign key for organism, Gabriel mentioned GNR, GPC, GPR so I'll change to use that as an example list
- [ ] we don't have a look up list for source, maybe use line source for the time being.

Other notes
I've gone for a vertical descending table, I don't think its a good idea to introduce horizontal scrolling.

I'm not sure about the fade to black on the loading between different screens, I like it for a long load, but not between for example lists and pathways, opinions and ideas welcome!

PR #79, opened 2016-09-07
remove additional options, move pathway into a single file, delete ad…

fredkingham:
…ditional models, add auto complete search


elcid/templates/pathway/cernerletter.html
```diff
@@ -1,6 +1,3 @@
-<hr />
-<h3>Clinical note on [[ editing.demographics.first_name ]] [[ editing.demographics.surname ]]</h3>
-
```
davidmiller, line 3: Are we deliberately removing this ?

fredkingham:
Yeah it was duplicated (http://elcid-rfh-test.openhealthcare.org.uk/pathway/#/cernerdemo) so we now just have it on line 211 on the pathway template

davidmiller:
Oh I see - great !

Hard to spot from the diff :)

PR #85, opened 2016-10-10
Pseudo hl7 results

fredkingham:


fredkingham:
Note the hardcoded results are commented out, I think its useful to keep them short term, but we'll remove them long term

PR #106, opened 2016-12-15
Gloss integration

fredkingham:


fredkingham:
The post save model hook didn't work because we save earlier than we update. Moved the gloss api logic into the pathway.

Moved the add patient form from cerner out of the navbar as discussed.

PR #110, opened 2017-01-13
Add server side gloss refresh

fredkingham:


davidmiller:
Are we sure that results won't have multiple entries for themselves ?
e.g. refresh updates not creates for existing tests

Do we want the endpoints to be part of the Opal api namespace ? 

Could we have a management command that runs refresh for all patients (in order to later cron it)

davidmiller:
Closing this for now to return and finish it in the next sprint - hopefully with credentials to the RFH archive server.

PR #134, opened 2017-02-08
mark an episode if they have a historically positive blood culture

fredkingham:


davidmiller:
Can we just check that we don't need to check for Tagging.archived in the save signal ? 
(I think we're fine, but would be neat to have a test for edge cases explicitly and a comment there would be awesome! )

PR #168, opened 2017-04-26
timeline clincial advice should be editable (timeline template part 2)

fredkingham:


fredkingham:
moving to merge into v0.2.3

PR #169, opened 2017-04-26
Add copy to clipboard directve  (timeline template part 3)

fredkingham:


fredkingham:
moving to depend on v0.2.3

PR #181, opened 2017-05-02
menu items fix backported from v0.2.3

fredkingham:


davidmiller:
I've checked out elCID@RFH v0.2.2.1 and Opal v0.8.1 and then added the extra tests from this PR, but they don't fail  - what should I be doing to test that and get a failure?

fredkingham:
hey, it was using the same reference, and there fore falling into the same trap, altered, try it now

PR #230, opened 2017-06-22
Lab tests demo

fredkingham:


davidmiller:
I might be reading Coveralls reports wrong, but I _think_ we still need some tests for `detail.py`, `api.py` and `elcid/assets/js/elcid/controllers/result_view.js` based on https://coveralls.io/builds/12103276 ?

davidmiller:
@fredkingham How relevant is this branch/PR now ?

PR #231, opened 2017-06-22
exclude lab test serialisation from the list view

fredkingham:


davidmiller:
I think we probably need to check that there's a tagged patient list in the serialiser tests?

And perhaps a test for `.to_dict()` that ensures we still only fetch episodes with the relevant tags?

davidmiller:
@fredkingham I don't think this can be targetting the right branch any more ?

PR #237, opened 2017-06-28
Hide toggle rfh list

fredkingham:


fredkingham:
This doesn't have animation, I need to look more closely at why animation is currently not working

PR #248, opened 2017-07-27
removes the episode comparators as the opal default should now be better

fredkingham:
Depends on https://github.com/openhealthcare/opal/pull/1211


fredkingham:
I don't think we want to do this because the default uses episode start and end, however start and end are dates whereas episode id gives us intra day ordering

PR #252, opened 2017-08-05
A custom find patient step for the royal free that doesn't create pat…

fredkingham:
…ients


davidmiller:
Looks good but can we have some comments in the bits that pull context from `$parent` ?

PR #261, opened 2017-09-18
Deployment script

fredkingham:


conf_templates/cron_copy.jinja2
```diff
@@ -0,0 +1 @@
+0 2 * * * {{ unix_user }} {{ virtualenv }}/bin/fab -f {{ fabric_file }} copy_backup:{{ branch }}
```
davidmiller: How do you know that this deterministically runs after today's database backup ?

Also, worth redirecting stdout and error to a log file...

conf_templates/gunicorn.conf.jinja2
```diff
@@ -3,7 +3,7 @@
 ; =====================
```
davidmiller: How come we're putting these in conf_templates as a top level directory?
We now have at the top level: 
./config
./conf_templates
./etc
./fabric_tests 

Maybe that's too many sysadminy / deploymenty directories at the top level ? :)

elcid/settings.py
```diff
@@ -180,7 +180,7 @@
 )
 
 if 'test' in sys.argv:
-    INSTALLED_APPS += ('opal.tests',)
+    INSTALLED_APPS += ('opal.tests', 'fabric_tests',)
```
davidmiller: Why don't these go in ./elcid/test/ or ./elcid/test/fabric ?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+from fabric.context_managers import lcd, settings
+from fabric.decorators import task
+import os.path
+
+env.hosts = ['127.0.0.1']
+UNIX_USER = "ohc"
+DB_USER = "ohc"
+RELEASE_NAME = "elcidrfh-{branch}"
+
+VIRTUAL_ENV_PATH = "/home/{usr}/.virtualenvs/{release_name}"
```
davidmiller: Can we name the inline variable `unix_user` like the one below?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
-            # heroku somtimes has memory issues doing migrate
-            # it seems to work fine if we just migrate opal first
-            # it will later fail because content types haven't
-            # been migrated, but that's fine we'll do that later
-            local("python manage.py migrate opal")
-
-        for db_command in db_commands(username):
-            local(db_command)
+def create_empty_env(new_branch):
+    _deploy(new_branch)
```
davidmiller: Won't this now also restart the supervisord/gunicorn processes in the new environment if it just calls _deploy?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+            current_env
+        )
+    else:
+        with settings(warn_only=True):
+            failed = put(
+                local_path=current_env.backup_name,
+                remote_path=current_env.backup_name
+            ).failed
+
+        if failed:
```
davidmiller: Wouldn't we like to capture the nature/output of the failure?
I suspect emitting a timestamped log message of success that the cron job puts somewhere would also be useful when we come to debug these

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+            "unable to find backup {}".format(
+                current_env.backup_name
+            ),
+            current_env
+        )
+    else:
+        with settings(warn_only=True):
+            failed = put(
+                local_path=current_env.backup_name,
+                remote_path=current_env.backup_name
```
davidmiller: If we're putting the elcid0.2.1 backup on the test server in the same location, if elcid0.2.1 has been deployed to the test server with deploy_prod, how do we know that the live backup won't be overwritten? Can we not put it in a place that is explicitly live backups which are different from test backups?

conf_templates/cron_backup.jinja2
```diff
@@ -0,0 +1 @@
+0 2 * * * postgres pg_dump {{ database_name }} > {{ backup_dir }}/back.$(date +"\%d.\%m.\%Y").{{ database_name }}.sql
```
davidmiller: I think it might be worth redirecting stderr to a log file somewhere ?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+            branch_env.project_directory
+        )
+    )
 
 
 @task
-def checkout_project():
-    checkout(get_requirements())
+def create_private_settings():
+    with open(PRIVATE_SETTINGS, "w") as privado:
```
davidmiller: If this file already exists is the right thing to do definitely to overwrite it rather than error?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+        some_env
+    )
+
+
+@task
+def copy_backup(branch_name):
+    current_env = Env(branch_name)
+    private_settings = get_private_settings()
+    env.host_string = private_settings["host_string"]
+    env.password = private_settings["password"]
```
davidmiller: Shouldn't this setting be created with create_private_settings?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+
+    # django setup
+    run_management_command("collectstatic --noinput", new_env)
+    run_management_command("migrate", new_env)
+    run_management_command("load_lookup_lists", new_env)
+    restart_supervisord(new_env)
+    restart_nginx()
+
+
+def infer_current_branch():
```
davidmiller, line 564: Why don't we ask Git what the branch is ?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+def _deploy(new_branch, backup_name=None):
+    if backup_name and not os.path.isfile(backup_name):
+        raise ValueError("unable to find backup {}".format(backup_name))
+
+    # the new env that is going to be live
+    new_env = Env(new_branch)
+
+    # the private settings
+    private_settings = get_private_settings()
+    env.host_string = private_settings["host_string"]
```
davidmiller: What is this for ? I can't see a code path that uses this ?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+        )
+        result = local(cmd, capture=True)
+    return result
 
-    with lcd(fabfile_dir):
-        requirements = local("less requirements.txt", capture=True).split("\n")
 
-        package_to_version = {}
+def pip_create_virtual_env(new_env):
+    local("rm -rf {}".format(new_env.virtual_env_path))
```
davidmiller: We definitely want to always trash an existing virtualenv if we re-deploy a version?
What if it's during a deploy_prod call ?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+    return
 
-        for requirement in requirements:
-            parsed_url = parse_github_urls(requirement)
 
-            if parsed_url:
-                package_to_version.update(parsed_url)
+def pip_install_requirements(new_env, proxy):
+    pip = "{}/bin/pip".format(new_env.virtual_env_path)
+    local("{0} install --upgrade pip --proxy {1}".format(pip, proxy))
```
davidmiller: Do we definitely want to upgrade pip every time we deploy? What if there's a release after we deploy to a test server and before we deploy to prod that removes a `--option` we're using?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
-    python_cmd += "' | python ./manage.py shell"
-    cmds.append(python_cmd)
-    return cmds
-
-def push_to_heroku(remote_name):
-    with lcd(fabfile_dir):
-        current_branch_name = local(
-            "git rev-parse --abbrev-ref HEAD", capture=True
+    #  https://stackoverflow.com/questions/14549270/check-if-database-exists-in-postgresql-using-shell
+    command = "sudo -u postgres psql -lqt | cut -d \| -f 1 | grep -w {database_name} | wc -l"
```
davidmiller: This feels a bit `from stackoverflow import database_exists`...
(https://github.com/drathier/stack-overflow-import) I'm not sure I understand why this works..

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+    database_exists = "0" not in local(
+        command.format(database_name=some_env.database_name), capture=True
+    )
+
+    if database_exists:
```
davidmiller: If this was true during `deploy_prod` do we definitely want to drop the database?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+        postgres_command("DROP DATABASE {0}".format(some_env.database_name))
+
+    postgres_command("CREATE DATABASE {0}".format(some_env.database_name))
+    postgres_command("GRANT ALL PRIVILEGES ON DATABASE {0} TO {1}".format(
+        some_env.database_name, DB_USER
+    ))
+
+
+def postgres_load_database(backup_name, new_env):
+    if not os.path.isfile(backup_name):
```
davidmiller: Don't we also check that this file exists at the start of  `_deploy` ?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+    #  https://stackoverflow.com/questions/14549270/check-if-database-exists-in-postgresql-using-shell
+    command = "sudo -u postgres psql -lqt | cut -d \| -f 1 | grep -w {database_name} | wc -l"
+    database_exists = "0" not in local(
+        command.format(database_name=some_env.database_name), capture=True
+    )
+
+    if database_exists:
+        postgres_command("DROP DATABASE {0}".format(some_env.database_name))
+
+    postgres_command("CREATE DATABASE {0}".format(some_env.database_name))
```
davidmiller: Why do we sometimes use this and sometimes type `sudo -u postgres` etc into a string like e.g. L188 ?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+    local('sudo ln -s {0} {1}'.format(
+        abs_address,
+        symlink_name
+    ))
 
-@task
-def create_heroku_instance(name, username):
+
+def services_create_local_settings(new_env, additional_settings):
+    new_settings = copy.copy(additional_settings)
```
davidmiller: The fact that the settings dictionary in ../private_settings needs strings that include string quotes is going to catch me out in 6 months time...

conf_templates/upstart.conf.jinja2
```diff
@@ -26,4 +26,4 @@ pre-start script
     # echo "Put bash code here"
 end script
 
-exec su -c ". /home/ohc/.virtualenvs/elcid-rfh010/bin/activate; /home/ohc/.virtualenvs/elcid-rfh010/bin/supervisord -c /usr/local/ohc/elcid-rfh010/etc/production.conf >> /tmp/elcid.upstart.log 2>&1" ohc
+exec su -c ". {{ env_name }}/bin/activate; {{ env_name }}/bin/supervisord -c {{ project_directory }}/etc/production.conf >> /tmp/elcid.upstart.log 2>&1" ohc
```
davidmiller: I think the log file this is nominally redirecting to is probably wrong?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+def services_create_gunicorn_conf(new_env):
+    template = jinja_env.get_template('conf_templates/gunicorn.conf.jinja2')
+    output = template.render(
+        env_name=new_env.virtual_env_path
+    )
+    gunicorn_conf = '{0}/etc/gunicorn.conf'.format(
+        new_env.project_directory
+    )
+
+    local("rm -f {}".format(gunicorn_conf))
```
davidmiller: If this actually finds something, at least I'd want a log message?
I suspect maybe moving to e.g. ./etc/gunicorn.conf.bak might be worth doing? (I'd want to know what it was and if it had changed I imagine)

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+    template = jinja_env.get_template('conf_templates/upstart.conf.jinja2')
+    output = template.render(
+        env_name=new_env.virtual_env_path,
+        project_directory=new_env.project_directory
+    )
+    upstart_conf = '{0}/etc/upstart.conf'.format(
+        new_env.project_directory
+    )
+
+    local("rm -f {}".format(upstart_conf))
```
davidmiller: Similar to the rm -rf 'ing of gunicorn confs above...

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+    if not os.path.isfile(abs_address):
+        raise ValueError(
+            "we expect an nginx conf to exist at {}".format(abs_address)
         )
-        local("git push {0} {1}:master".format(
-            remote_name, current_branch_name)
+
+    symlink_name = '/etc/nginx/sites-enabled/{}'.format(MODULE_NAME)
+    if os.path.islink(symlink_name):
+        local("sudo rm {}".format(symlink_name))
```
davidmiller: I think I want a log/print here...

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+
+
+def services_symlink_upstart(new_env):
+    abs_address = "{}/etc/upstart.conf".format(new_env.project_directory, "")
+    if not os.path.isfile(abs_address):
+        raise ValueError(
+            "we expect an upstart conf to exist {}".format(abs_address)
         )
+    symlink_name = '/etc/init/{}.conf'.format(PROJECT_NAME)
+    local("sudo rm -f {}".format(symlink_name))
```
davidmiller: Why do we do this rather than checking if it's a link the way we do for nginx.conf?

(I think I also want a log/print here if we remove something)

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+    services_create_upstart_conf(new_env)
+
+    # symlink the nginx conf
+    services_symlink_nginx(new_env)
+
+    # symlink the upstart conf
+    services_symlink_upstart(new_env)
+
+    # django setup
+    run_management_command("collectstatic --noinput", new_env)
```
davidmiller, line 557: We don't need to check that these haven't failed?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
-    """
-    looks for a requirements file in the same directory as the
-    fabfile. Parses it,
-    """
+def run_management_command(some_command, env):
+    with lcd(env.project_directory):
+        cmd = "{0}/bin/python manage.py {1}".format(
+            env.virtual_env_path, some_command
+        )
+        result = local(cmd, capture=True)
```
davidmiller, line 141: Does capture=True mean that stdout/err are available on `result` ? Do they also get printed to the terminal?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+        local("pkill super; pkill gunic; pkill gloss_flask")
+    # don't restart supervisorctl as we need to be running the correct
+    # supervisord
+    local("{0}/bin/supervisord -c {1}/etc/production.conf".format(
+        new_env.virtual_env_path, new_env.project_directory
+    ))
+
+
+def restart_nginx():
+    local('sudo /etc/init.d/nginx restart')
```
davidmiller: Not `service nginx` ?

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+
+    # the new env that is going to be live
+    new_env = Env(new_branch)
+
+    # the private settings
+    private_settings = get_private_settings()
+    env.host_string = private_settings["host_string"]
+
+    # Setup environment
+    pip_create_virtual_env(new_env)
```
davidmiller, line 533: I think I want print to stdout as we perform each of these tasks somehow - either here or inside the functions...

fabfile.py
```diff
@@ -1,197 +1,518 @@
...
+    dump_str = "sudo -u postgres pg_dump {0} -U postgres > {1}"
+    local(dump_str.format(old_env.database_name, old_env.release_backup_name))
+
+    cron_write_backup(new_env)
+    cron_copy_backup(new_env)
+    old_status = run_management_command("status_report", old_env)
+    _deploy(new_branch, old_env.release_backup_name)
+    new_status = run_management_command("status_report", new_env)
+
+    print("=" * 20)
```
davidmiller: Does this mean that we're expected to eyeball these?
Can't we have a compare function so that a computer can test for equivalence? :)

davidmiller (commented):
This looks like a lot of comments... (Because it is a lot ;) 

I think I'd quite like a 'rollback to branch' command that for an existing deployment environment will just make that the one actively deployed, stopping supervisor/gunicorn, starting the named one, and replacing cron, upstart etc... 

So that if I decide to abort a deployment I can do e.g. 
`fab rollback 0.4`

PR #282, opened 2017-10-30
V0.2.8 with opal 9

fredkingham:


fredkingham:
I'm afraid this is the God pull request of opal v0.9.0.

Things to note.

1. deployment: on a test build we now delete the existing directory where we're deploying to if it exists. We also run the create_singletons management command

2. start end, in 0.9.0 we move to a start and end on episode. These were no being populated before. We use the 'Location' model as this is added in the referral pathway. We also now set the start in that pathway.

3. the clinical advice inline issues. We now switch out the episode on the item, just before the save. This works because of https://github.com/openhealthcare/opal/pull/1276/files

4. Removing unused forms

elcid/assets/js/elcid/controllers/haem_view.js
```diff
@@ -33,7 +33,7 @@ angular.module('opal.controllers').controller('HaemView', function($scope){
   };
 
   vm.getEpisodeOrdering = function(episode){
-      var significantDate = episode.discharge_date || episode.date_of_episode || episode.date_of_admission;
+      var significantDate = episode.end || episode.start;
```
davidmiller, line 36: I thought we just didn't use this ?

requirements.txt
```diff
@@ -34,6 +34,6 @@ Fabric==1.14.0
 
 # Open Health Care packages on PyPi
 opal-observations==0.3.0
+-e git+https://github.com/openhealthcare/opal.git@search-seprate-with-9#egg=opal
```
davidmiller: I'm not sure we want to deploy/merge this ?

elcid/assets/js/elcid/controllers/clinical_advice_form.js
```diff
@@ -25,6 +25,7 @@ angular.module('opal.controllers').controller(
             self.save = function(){
               ngProgressLite.set(0);
               ngProgressLite.start();
+              item.episode = $scope.episode;
```
davidmiller: Does this work ? :D 

There are definitely no e.g. trailing other references to episode in closures inside item?

elcid/assets/js/elcid/controllers/haem_view.js
fredkingham: good point, we don't

requirements.txt
fredkingham: (yep, I'll push a fix)

elcid/assets/js/elcid/controllers/clinical_advice_form.js
fredkingham: I updated it so there weren't any, it is something that I was worried was fragile, but the unit tests cover it so if we change the opal side they will break.

fredkingham:
we expect unit tests to file until we've deployed to pypi

PR #284, opened 2017-10-30
only serialise specific models

fredkingham:


fredkingham:
coming in a different pull request

PR #310, opened 2017-12-03
make hospital number required in the reconciliation

fredkingham:
We should never be saving a patient without a hospital number, so lets make sure of that


fredkingham:
I'm not sure this is still needed

Issue #313, opened 2017-12-04
Change models to extend from opal core models.


fredkingham:
In Opal we have core models such as [Demographics](https://github.com/openhealthcare/opal/blob/v0.9.1/opal/models.py#L1374)

At the Royal Free do not seem to be using these models for [example](https://github.com/openhealthcare/elcid-rfh/blob/v0.2.9/elcid/models.py#L34)

- [ ] where the fields are the same, or the fields at the Royal free are a superset, change to use the opal core models.
- [ ] create and run migrations

pacharanero:
@davidmiller and I paired on this today and we hit issues refactoring the Allergies model to use opal.core.models.Drug, which for reasons of time has been left as a WIP in the https://github.com/openhealthcare/elcid-rfh/tree/monster-refactoring-of-antimicrobial-class branch

pacharanero:
We think it might be better to complete #315 first before coming back to this - there's no point refactoring models that aren't used :-)

fredkingham:
so it turns out this really wan't easy!

So I've extended diagnosis.

What would you say to moving the Antimicrobial model into elcid and then renaming the model in opal to Drug?

Thoughts @davidmiller

If we agree, I'll close this ticket and open a new one.

I now disagree with this, I'm going to create a ticket that does what you and @pacharanero started.

Current theory, rename the Antimicrobial method in Opal to Drug. In many ways this makes more sense. In practice it might be too complicated to coordinate.

fredkingham:
closing this, we need to work out if Antimicribial == Drug, but I think this is for another time.

PR #322, opened 2017-12-11
changed demographics model, migration

pacharanero:
This PR addresses the Demographic model in elCID as per #313


davidmiller:
I don't think we should worry too much about falling coverage - this is just removing code...

PR #325, opened 2017-12-12
314 update fields to use form api

### LINKED ISSUE #314: update fields to use the form api
some forms are not using the field api
- [ ] elcid/templates/forms/primary_diagnosis_form.html
- [ ] elcid/templates/forms/primary_diagnosis_form.html
- [ ] elcid/templates/forms/line_form.html
- [ ] elcid/templates/forms/antimicrobial_form.html
- [ ] elcid/templates/forms/imaging_form.html

pacharanero:
closes #314 

note some tweaks to the default names of inputs for clarity, eg distinguishing between date/time of Insertion and date/time of removal (which are both on the same form)

image


fredkingham:
love the datetime change, I think that's great.

PR #330, opened 2017-12-20
changes date ordered to datetime ordered

fredkingham:


fredkingham:
should pass after https://github.com/openhealthcare/elcid-rfh/pull/331 is merged

PR #334, opened 2018-01-02
Update 500 to use static_page.html

davidmiller:


fredkingham:
It looks like 500 pages don't get run through the template context processors.

When we try and get menu items we require 'user' so it blows up and we never hit the 500.

I think this is more an issue with menus...

davidmiller:
Is that comment meant to go here? 
Isn't that what https://github.com/openhealthcare/opal/pull/1338 is about or a different issue?

fredkingham:
excellent! I hadn't seen that pull request 👍

PR #335, opened 2018-01-02
bumps pyscopg2

fredkingham:


davidmiller:
I'm not entirely sure that a PR for just this version bump is the right thing to do as opposed to upgrading to a 10 branch, but sure...

PR #346, opened 2018-01-25
creates a couple of patient lists PCP and R1

fredkingham:


fredkingham:
This has been released

PR #349, opened 2018-01-29
remove the demographic detail it does not make any sense and we don't…

fredkingham:
… use it.

It looks like its pretty malformed. Its missing a closing div and the last line is just a bit wrong.


davidmiller:
Stale target branch now - moving to #482

Issue #352, opened 2018-02-08
replace the referral panel in the initial assessment form


fredkingham:
Change it to a drop down list of what is in the printed out Form 1.

They only care about date, about that list.
We should store it in referral type.

davidmiller:
Closing as PR merged already

PR #364, opened 2018-02-08
changes it to a select box, because our radio button formatting is a …

fredkingham:
…bit off for this amount of options. (also we'd like to use bootstrap buttons anyway). We need to delete existing lookup lists before deployment


fredkingham:
fixes #352

PR #368, opened 2018-02-08
lets use the same requests as opal, fixes travis

fredkingham:


fredkingham:
@davidmiller I am merging this because I can't think this would have issues and it means we can fix it up for the other pull requests that are many.

PR #371, opened 2018-02-12
adds in the new patient lists R1, PCP and Liver Transplantation

fredkingham:


fredkingham:
fails because of pyscopg2. we've fixed this in 3, I'd like to push forward without updating in this version, your call though

PR #372, opened 2018-03-08
Lab test async

fredkingham:
initial async that has been what we've presenting


fredkingham:
This is dead, its at the very least included in https://github.com/openhealthcare/elcid-rfh/pull/384

PR #376, opened 2018-04-04
Add alternative localhost spellings

davidmiller:


davidmiller:
Stale target branch - moving to #483

PR #394, opened 2018-05-09
change to use get params for the search api to allow for search param…

### LINKED ISSUE #386: Searching for hospital number including `.` results in an alert()
Maybe handle with either validation or just accept the search ?

fredkingham:
…s with punctuation

fixes #386


davidmiller:
I can't see the tests that test it actually fixes the issue ? 

For `.` `,` `/` etc etc?

Issue #391, opened 2018-04-26
Box heights jump around all over the place when loading trends


davidmiller:
Can we make that feel smoother with for instance a judicious `min-height` property somewhere?

PR #399, opened 2018-05-10
391 box heights

fredkingham:
Ummed and Arrred about whether to add this to the list page as well. I ended up adding it to both.

The toss up is whether you want the list view jumping on load vs whether you want large empty spaces if lab tests aren't loaded for a lot of patients.

Looking at tests the vast majority of patients have lab tests so I have erred on the side of adding a min height to trends on both screens.


fredkingham:
fixes #391

PR #407, opened 2018-05-14
We use the error emailer to send us an email if there is a js error, …

### LINKED ISSUE #406: Catch and post js errors
if there is a javascript error we should get an email. It should also be logged by the client

fredkingham:
…the details will be excluded from the email but contained in the console log fixes #406


elcid/__init__.py
```diff
@@ -25,12 +25,13 @@ class Application(application.OpalApplication):
         'js/elcid/controllers/remove_patient_step.js',
 
         'js/elcid/services/demographics_search.js',
+        'js/elcid/services/demographics_search.js',
 
 
         'js/elcid/controllers/tagging_step.js',
         'js/elcid/controllers/investigations.js',
         # used in the blood culture forms
```
davidmiller: I'm not sure the comment is useful when it's above exception_handler.js ....

elcid/assets/js/elcid/services/exception_handler.js
```diff
@@ -0,0 +1,13 @@
+angular.module('opal.services').factory('$exceptionHandler', function($injector, $log) {
+    "use strict";
+    var errorCatcherHandler = function errorCatcherHandler(exception, cause) {
+        // we use the injector to stop a circular dependency as per
+        // https://stackoverflow.com/questions/22332130/injecting-http-into-angular-factoryexceptionhandler-results-in-a-circular-de
+        var $http = $injector.get("$http");
+        var url = "/elcid/v0.1/error_emailer/"
+        $http.post(url, {exception: exception, cause: cause});
```
davidmiller, line 8: Do we have an illustrative value of the `cause` argument ?

davidmiller, line 8: How sure are we that an uncaught HTTP failure / 404/500 won't drop into an infinite loop ?

fredkingham: so a 500 doesn't cause an infinite loop. Even if the error emailer itself throws a 500 we don't go infinite. So I think we're good

elcid/__init__.py
fredkingham: 👍

elcid/assets/js/elcid/services/exception_handler.js
fredkingham:
```
<div class="outer-container ng-scope" ng-view="" autoscroll="true" ng-hide="ready==false">
```

ie it gives you the element from where you're declaring the scope that threw the error. Not really that useful for us, but I figured more data is better than less. If we ever move to having multiple angular apps it might be useful it could be handy.

davidmiller: What if there's an error in the call to $http.post somewhere though ?

fredkingham: It all I _think_ works fine, for example if you don't pass in a url to a get request it raises an error but only for that request, we don't tumble into an infinite loop. It all works as one would hope. ie pinging the error emailer etc.

davidmiller:
This looks like it could be a good idea... it catches errors on the front end and makes sure they're robustly logged and we hear about it?

Def. needs a new target branch...

fredkingham:
@davidmiller this is one of those changes that I really want, I can't see any problems with but I'm scared of.

I'd suggest putting it out on a release by itself rather than the next release just to minimise the effect of any unknown issue it might have.

Thoughts?

PR #409, opened 2018-05-15
run lab tests once every 6 mins not once an hour refs #408

### LINKED ISSUE #408: RFH Lab Tests are running once an hour rather than once every 5 mins
At the moment the batch patient load loads takes between 2 and 5 mins.

This is because we iterate through demographics for the non reconciled. The batch patient load takes significantly less.

I would suggest updating to run every 6 mins.

We can look at updating the demographics logic at a different point.

fredkingham:
fixes #408


davidmiller:
@fredkingham I think this is now about to be superseded ?

fredkingham:
yep, not needed anymore

PR #418, opened 2018-05-24
remove the ipl from the batch synch wait

fredkingham:


fredkingham:
closing, this as we are looking at resolving it in another way.

PR #419, opened 2018-06-04
414 fixes 1

fredkingham:


elcid/templates/partials/lab_test_sparklines.html
```diff
@@ -11,7 +15,18 @@ <h2 class="text-center no-results">Unable to get lab tests for this patient</h2>
     <h2 class="text-center no-results">Failed To load lab tests for this patient</h2>
   </div>
   <div ng-show="patientLoadStatus.isLoaded()">
-    <div ng-hide="data.obs_values.length">
+    {% comment %}
+      we have a result saying that the patients lab tests have loaded but
+      we have not loaded them in yet
```
davidmiller, line 20: I'm not sure I understand this comment ? You mean 'we think there are tests, this only displays when we haven't fetched them from the API yet' ?

etc/conf_templates/cron_lab_tests.jinja2
```diff
@@ -1 +1 @@
-0/5 * * * * {{ unix_user }} {{ virtualenv }}/bin/python {{ project_dir }}/manage.py batch_load >> /usr/lib/ohc/log/cron_synch.log 2>&1
+*/15 * * * * {{ unix_user }} {{ virtualenv }}/bin/python {{ project_dir }}/manage.py batch_load >> /usr/lib/ohc/log/cron_synch.log 2>&1
```
davidmiller, line 1: This isn't what I was expecting...
According to this website this is 'load every 15 mins instead of every 5'

New:
https://cronexpressiondescriptor.azurewebsites.net/?expression=*%2F15+*+*+*+*&locale=en
Old:
https://cronexpressiondescriptor.azurewebsites.net/?expression=0%2F5+*+*+*+*+&locale=en

intrahospital_api/apis/prod_api.py
```diff
@@ -146,6 +143,20 @@ def get_or_fallback(self, primary_field, secondary_field):
 
         return result
 
+    def get_identifier(self):
+        """ for reasons unknown the upstream database often has the
```
davidmiller, line 147: It would be interesting for this to be less unkonwn... :)

etc/conf_templates/cron_lab_tests.jinja2
fredkingham: I think that might be true for Azure but not for traditional linux. Scary that it appears highly on cron descriptors google search.

I for one am slightly wary of our new github overlords.

intrahospital_api/apis/prod_api.py
fredkingham: I was assuming it was for patients not at the RFH, ie liaison, but some investigation could be done.

elcid/templates/partials/lab_test_sparklines.html
fredkingham: Maybe I can explain it better..

Yeah we've validated that tests have loaded, but we have not loaded them in yet.

fredkingham:
closing this as its been superseded by other PRs

Issue #420, opened 2018-06-14
the back up script should blow up if it cannot run for an hour


fredkingham:
if batch loads or ipls are running the back up script waits.

If it is waiting for more than an hour the process should die and send an email.

davidmiller:
@fredkingham Is this done?

fredkingham:
The back up script no longer waits for batch loads to be completed. This resolves the issue.

PR #421, opened 2018-06-14
if we have any issues with the ack up script, don't wait indefinitely…

fredkingham:
…, wait it out, but then email us refs #420

### DIFF
--- a/elcid/test/test_fabric.py
+++ b/elcid/test/test_fabric.py
@@ -1300,6 +1300,99 @@ class DiffStatusTestCase(FabfileTestCase):
         )
 
 
+@mock.patch("fabfile.dump_database")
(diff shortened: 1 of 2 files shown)


elcid/test/test_fabric.py
```diff
@@ -96,6 +96,7 @@ def test_private_settings_already_exist(self, os, json):
 
 class FabfileTestCase(OpalTestCase):
     def setUp(self):
+        print "starting runing for a test for {}".format(self.__class__)
```
davidmiller: Parentheses for print - we should be aiming for python3 compatible where possible I think....

fredkingham: Great spot, this should not have been there, removed

PR #457, opened 2018-08-06
1573 pathway step save

fredkingham:
depends on openhealthcare/opal#1573


davidmiller:
Do we still want this ?
If so, tests need to pass, and it's dependent on an open PR

3.5 certainly not the right branch for this to go into obv.

PR #484, opened 2018-09-27
With added organisms

fredkingham:


davidmiller:
@fredkingham What happens if we merge this and then run `load_lookup_lists` as part of a deployment? Will it do the right thing? Have we tested it?

fredkingham:
@davidmiller yep, tested and it adds in the expected organism and does not remove any.

davidmiller:
Merged into 0.3.7 on the commandline, closing this.

PR #485, opened 2018-09-27
Growl alert background color

fredkingham:


fredkingham:
screen shot 2018-09-27 at 16 20 45

davidmiller:
I think this needs merge conflicts resolving now...
If we're going to do it, can we not find a css class for the tick and make it white?

fredkingham:
@davidmiller its not a css class its an image 😞 I can try to colour it but it'll be awkward...

fredkingham:
The previous behaviour is preferred.

davidmiller:
screen shot 2018-10-02 at 15 58 44
Honestly, I prefer the green on green to green on blue if we can't easily change the image....

Issue #487, opened 2018-10-01
TB Phase 0


davidmiller:
- [x] Hide lists menu button
- [x] Failed search to go to the right add patient screen / steps 
- [x] Pathway title to not say TB
- [x] Hide appointments panel
- [x] Borders on print advice: Make them greyscale by default
- [x] Add an icon to imaging

fredkingham:
- [ ] Add a lookup list  of `state` which is a proxy for stage at this time, populated from a lookup list

davidmiller:
This is done and deployed, closing.

PR #506, opened 2018-10-02
adds in tb stage as a subrecord refs #487

fredkingham:


apps/tb/models.py
```diff
@@ -448,7 +448,46 @@ class Meta:
...
     case_manager = ForeignKeyOrFreeText(TBCaseManager)
-    ltbr_number  = fields.CharField(
+    ltbr_number = fields.CharField(
         max_length=200, blank=True, null=True,
-        verbose_name = "LTBR Number"
+        verbose_name="LTBR Number"
     )
+
+
+class TBAppointment(models.PatientSubrecord):
```
davidmiller, line 457: I don't think we want this here?

davidmiller:
I think actually it might be best to hold out on these until we've looked in person... I think we need to get the lookuplist values right as currently used.

Also, Stage is going to be a confusing name given `EpisodeCategory.stages`

fredkingham:
We don't want to do this

Issue #514, opened 2018-10-05
Antimicrobials has been renamed Medication History


fredkingham:
In v0.3.6 the display name of Antimicrobial seems to have been accidentally renamed (a la `verbose_name`) to Medication History.

This is not expected should be rolled back.

PR #515, opened 2018-10-05
renames antimicrobials to be medication history in the tb record pane…

fredkingham:
…l only refs #514


fredkingham:
merged as agreed with @davidmiller

PR #518, opened 2018-10-09
aligns the print advice headings

fredkingham:


fredkingham:
screen shot 2018-10-09 at 12 26 59

PR #528, opened 2018-10-10
V0.3.7 api refactor

fredkingham:
So this is the first (and the worst) of the api refactor pull requests.

Old state was 
* An API that provided all database access.
* A loader file that would get the data from the api and pass it to updater files, e.g. update_lab_tests
* There was a single batch load cron job that would run every hour.

New State
* There are many services, each with their own cron jobs, backend and service.
* The `base` service provides utilities to help the services.
* Each service defines a `/backend/` directory and a live.API and a dev.API that is used depending on a setting.
* Each service has a `service.py` that requests data from the api and saves.
* There will be one batch loads for _each_ service.


fabfile.py
```diff
@@ -228,6 +227,27 @@ def postgres_load_database(backup_name, new_env):
     ))
 
 
+def write_cron_backup(new_env):
```
davidmiller, line 230: @fredkingham This function has just moved and had a docstring line added?

fabfile.py
```diff
@@ -384,41 +404,25 @@ def restart_nginx():
...
-        unix_user=UNIX_USER
-    )
-    cron_file = "/etc/cron.d/{0}_backup".format(PROJECT_NAME)
-    local("echo '{0}' | sudo tee {1}".format(
-        output, cron_file
-    ))
+def write_cron_jobs(new_env):
+    template_dir = "etc/cron_templates"
 
+    cron_jobs = os.listdir(template_dir)
```
davidmiller: I would probably have done a glob match on e.g. template_dir + `*.jinja2` to avoid the case where for example a `.swp` file gets made a cron job by accident and we run everything twice at the same time.

intrahospital_api/loader.py
```diff
@@ -49,85 +45,46 @@
...
-from intrahospital_api import get_api
 from intrahospital_api.exceptions import BatchLoadError
 from intrahospital_api.constants import EXTERNAL_SYSTEM
 from intrahospital_api import update_demographics
-from intrahospital_api import update_lab_tests
+from intrahospital_api.services.lab_tests import service as lab_tests
 from intrahospital_api import logger
-
-api = get_api()
+from intrahospital_api import get_api
```
davidmiller: Can we have all the `from intrahospital_api import x` statements on one line ?

intrahospital_api/loader.py
```diff
@@ -49,85 +45,46 @@
...
 
 @timing
-def initial_load():
-    models.InitialPatientLoad.objects.all().delete()
-    models.BatchPatientLoad.objects.all().delete()
-    batch = models.BatchPatientLoad()
-    batch.start()
+def initial_load(remaining=False):
+    """
+    Runs an initial load.
```
davidmiller, line 62: This isn't a great docstring comment - it assumes I know what running an  initial load is, when I would do it etc...

intrahospital_api/models.py
```diff
@@ -107,17 +107,26 @@ class BatchPatientLoad(PatientLoad):
         every 5 mins
```
davidmiller: Is this docstring inaccurate now ?

intrahospital_api/services/base/db.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    FIELD_MAPPINGS = {}
+
+    def __init__(self, db_row):
+        self.db_row = db_row
+
+    def get_or_fallback(self, primary_field, secondary_field):
+        """ look at one field, if its empty, use a different field
+        """
+        # we use Cerner information if it exists, otherwise
```
davidmiller: It's not really specific to Cerner I don't think?

intrahospital_api/services/base/db.py
```diff
@@ -0,0 +1,136 @@
...
+
+    db_row = dict(
+        patient_number="1",
+        first_name="Wilma",
+        last_name="Flintstone",
+        secondary_field="1/1/2000"
+    )
+    demographics_row = DemographicsRow(db_row)
+
+    demographics_row["hospital_number"] == "1"
```
davidmiller: You prefer this rather than for instance `demographics_row.hospital_number` ?

intrahospital_api/services/base/db.py
```diff
@@ -0,0 +1,136 @@
...
+    demographics_row["name"] == "Wilma Flintstone"
+    demographics_row["date_of_birth"] == "1/1/2000"
+    """
+    FIELD_MAPPINGS = {}
+
+    def __init__(self, db_row):
+        self.db_row = db_row
```
davidmiller: Bikeshedding alert, but given that the whole class is representing a database row, I'd probably avoid `Row.db_row`... something like `raw_data` ?

intrahospital_api/services/base/db.py
```diff
@@ -0,0 +1,136 @@
...
+    db_row = dict(
+        patient_number="1",
+        first_name="Wilma",
+        last_name="Flintstone",
+        secondary_field="1/1/2000"
+    )
+    demographics_row = DemographicsRow(db_row)
+
+    demographics_row["hospital_number"] == "1"
+    demographics_row["name"] == "Wilma Flintstone"
```
davidmiller: I definitely think it's weird to access `@property`s this way...

intrahospital_api/services/base/db.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if not result:
+            result = self.db_row.get(secondary_field, "")
+
+        return result
+
+    def __getitem__(self, key):
+        translated_field = self.FIELD_MAPPINGS[key]
+
+        if isinstance(translated_field, (tuple, list)):
```
davidmiller, line 102: I'd be tempted to have this logic in the `get_or_fallback` ?

elcid/settings.py
```diff
@@ -196,7 +200,6 @@
     database=None,
```
davidmiller: FWIW, Django tends to do these as shouty ALL_CAPS keys....

intrahospital_api/services/base/db.py
```diff
@@ -0,0 +1,136 @@
...
+                translated_field[0], translated_field[1]
+            )
+
+        if hasattr(self, translated_field):
+            return getattr(self, translated_field)
+        return self.db_row[translated_field]
+
+
+class DBConnection(object):
+    IP_ADDRESS = settings.HOSPITAL_DB.get("ip_address")
```
davidmiller: If we do these as .get() it means you never raise an error for missing one....

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    batch_loads = models.BatchPatientLoad.objects.filter(
+        service_name=service_name
+    )
+
+    if not batch_loads.exists():
+        return True
+
+    current_running = batch_loads.filter(
+        Q(stopped=None) | Q(state=models.BatchPatientLoad.RUNNING)
```
davidmiller, line 43: I'm not sure what SQL this outputs ? I don't understand bitwise operator overloading on `Q` objects :(

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if current_running.count() > 1:
+        # we should never have multiple batches running at the same time
+        raise BatchLoadError(
+            "We appear to have {} concurrent batch loads".format(
+                current_running.count()
+            )
+        )
+
+    last_run_running = current_running.last()
```
davidmiller: Why `last()` here?
We know there is one only - when I see `queryset.last()` I intuitively think it's because there might be > 1 ...

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if last_run_running:
```
davidmiller, line 47: What does this if statement check ?

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+        time_ago = timezone.now() - last_run_running.started
+
+        if time_ago.seconds < MAX_ALLOWABLE_BATCH_RUN_TIME:
+            logger.info(
+                "batch still running after {} seconds, skipping".format(
```
davidmiller: What do we "skip" at the point of this log message?

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+                    time_ago.seconds
+                )
+            )
+            return False
+        else:
+            logger.error(
```
davidmiller, line 59: Does this trigger an email?

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+            return False
+
+    return True
+
+
+def get_batch_start_time(service_name):
+    """
+    If no previous successful batch run has started.
+    Use the started timestamp of the first Initial Patient Load
```
davidmiller, line 72: First ? I thought there was never > 1 ?

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+    if not batch_loads.exists():
+        return True
+
+    current_running = batch_loads.filter(
+        Q(stopped=None) | Q(state=models.BatchPatientLoad.RUNNING)
+    )
+
+    if current_running.count() > 1:
+        # we should never have multiple batches running at the same time
+        raise BatchLoadError(
```
davidmiller: You prefer to do your own checks and raise your own errors not have e.g. a Meta unique_together constraint ?

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+    return True
+
+
+def get_batch_start_time(service_name):
+    """
+    If no previous successful batch run has started.
+    Use the started timestamp of the first Initial Patient Load
+
+    If a previous batch load has run and there was an initial patient
```
davidmiller, line 74: Can we have a diagram that explains this ? 
I'm only 75% sure I follow the text :)

fabfile.py
fredkingham: yes its exactly the same.

fabfile.py
fredkingham: 👍

intrahospital_api/services/base/db.py
fredkingham: 👍

intrahospital_api/services/base/db.py
fredkingham: 👍

intrahospital_api/services/base/db.py
fredkingham: 👍

intrahospital_api/services/base/load_utils.py
fredkingham: Its an SQL `or`, its not too exciting its just django having a weird api on it

intrahospital_api/loader.py
```diff
@@ -367,40 +130,31 @@ def async_load_patient(patient_id, patient_load_id):
...
 
+@timing
+def load_demographics(hospital_number):
+    started = timezone.now()
+    api = get_api()
+    try:
+        result = api.demographics(hospital_number)
+    except:
+        stopped = timezone.now()
+        logger.info("demographics load failed in {}".format(
```
davidmiller: Wouldn't you want to know demographics load for `hospital_number` ?

intrahospital_api/loader.py
```diff
@@ -367,40 +130,31 @@ def async_load_patient(patient_id, patient_load_id):
         raise
 
 
+@timing
+def load_demographics(hospital_number):
+    started = timezone.now()
+    api = get_api()
+    try:
+        result = api.demographics(hospital_number)
+    except:
```
davidmiller: Why the bare except here?

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+    if long_patient_load:
+        return long_patient_load.started
+    else:
+        return last_successful_run.started
+
+
+def batch_load(service_name):
+    """
+    A decorator that runs the function as batch load
+    ie wrapping it in the batch load decorator
```
davidmiller, line 133: I'm not really sure what this docstring is trying to say ?

e.g. I know it's a decorator because it has `@wraps` and looks like a decorator?

What does the decorator do?
When do we use it?

It reads like an explanation that assumes I already know what the thing we're explaining is...

intrahospital_api/services/base/load_utils.py
```diff
@@ -0,0 +1,160 @@
...
+        def wrap(*args, **kwargs):
+            if not good_to_go(service_name):
+                logger.info("batch {} not ready skipping".format(service_name))
+                return
+            logger.info("starting batch load for {}".format(service_name))
+            batch = models.BatchPatientLoad(service_name=service_name)
+            batch.start()
+            try:
+                count = fun(*args, **kwargs)
+            except Exception as e:
```
davidmiller, line 158: Why `Exception as e` syntax if we don't use `e` ?

intrahospital_api/management/commands/batch_test_load.py
```diff
@@ -0,0 +1,14 @@
+"""
+A management command that runs every hour and loads in
+lab tests of patients
+"""
+from django.core.management.base import BaseCommand
+from intrahospital_api.services.lab_tests import service
+
+
+class Command(BaseCommand):
+    help = " runs a batch load since the last successful batch run. --force\
```
davidmiller: I think this help string is wrong ? (In fact I think `--force` would raise some sort of exception?)

Also, while we're looking at it, sentence case for sentences...

intrahospital_api/services/base/load_utils.py
fredkingham: yep

intrahospital_api/services/base/load_utils.py
fredkingham: So its not unique together on state as we should have lots of batch runs that are successful. However you're right, we don't want to blow up here, we want to blow up when duplicates are added. I'll put an error in the save method

intrahospital_api/loader.py
fredkingham: There is not necessarily a demographics load. This is the function that's called by the add patient pathway.

intrahospital_api/loader.py
fredkingham: To stop it blowing up in the pathway.

PR #530, opened 2018-10-11
526 deployment check port

fredkingham:


intrahospital_api/deployment_check.py
```diff
@@ -0,0 +1,201 @@
+"""
+Gets called by the deployment check management command.
+
+check_since is the main method
+
+Takes a datetime and looks at all patients who don't have archived tags.
+
+Looks at all lab tests that have a last_updated after the datetime for those patients.
```
davidmiller, line 8: Can we make explicit what 'looks at' means - the phrase is used 3 times in 3 lines, but I still don't know what this 'looks at' or what question about these things we answer by calling `check_since()`.

intrahospital_api/deployment_check.py
```diff
@@ -0,0 +1,201 @@
...
+
+Then deletes them and runs a batch load since that time. Looks again.
+
+Then deletes them and runs initial patient loads for all patients. Looks again.
+
+The rolls back all changes in a transaction.
+
+It populates a result dict with
+current: The current state of the db.
```
davidmiller, line 17: Can we be more explicit than 'current state' ? e.g. I don't know what that means from this text, I have to read it, run the code, or both to find out...

intrahospital_api/deployment_check.py
```diff
@@ -0,0 +1,201 @@
...
+from opal.models import Patient, Tagging
+from elcid import patient_lists
+from lab import models as lmodels
+from intrahospital_api import models as imodels
+from intrahospital_api import loader
+from intrahospital_api import get_api
+from intrahospital_api import update_lab_tests
+
+
+class RollBackError(Exception):
```
davidmiller, line 41: Does elCID not have it's own `exceptions` module?

intrahospital_api/management/commands/deployment_check.py
```diff
@@ -0,0 +1,42 @@
...
+
+        since = timezone.now() - datetime.timedelta(
+            seconds=options["hours"] * 60 * 60
+        )
+
+        try:
+            deployment_check.check_since(since, result)
+        except deployment_check.RollBackError:
+            six._print(json.dumps(result, indent=2))
+        except Exception as e:
```
davidmiller, line 40: Are you sure we want to catchall and then `pass` ? That seems... aggressive and hard to debug?

intrahospital_api/management/commands/deployment_check.py
```diff
@@ -0,0 +1,42 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'hours',
+            help="hours ago",
+            type=int,
+            default=4
+        )
+
+    def handle(self, *args, **options):
+        result = {}
```
davidmiller, line 30: I'm not really sure why we're passing in this dictionary ? Can't we just return from a function ?

intrahospital_api/deployment_check.py
```diff
@@ -0,0 +1,201 @@
...
+
+    the result are observations, lab number, patient id that do not exist at
+    least 1 other state (be it current, post batch load or post initial load)
+
+    the count is the total observations loaded regardless of whether
+    they are in another load or not
+    """
+
+    if "test" not in settings.OPAL_BRAND_NAME.lower():
+        raise ValueError('this should only be run on a test server')
```
davidmiller, line 171: This just gets swallowed by the parent catchall ATM though ?

intrahospital_api/deployment_check.py
```diff
@@ -0,0 +1,201 @@
...
+
+    current, max_updated = get_and_clean(some_dt)
+    result["current_count"] = len(current)
+
+    patients = get_patients()
+    batch_load_patients(some_dt)
+    batch_load, _ = get_and_clean(some_dt, max_updated)
```
davidmiller, line 178: Is there a reason for implicit keyword arguments rather than explicit?

intrahospital_api/deployment_check.py
```diff
@@ -0,0 +1,201 @@
...
+def get_key(observation_number, lab_test):
+    """
+    A key is the observation number, lab test number and patient id.
+    """
+    return (
+        observation_number, lab_test.external_identifier, lab_test.patient_id,
+    )
+
+
+def get_and_clean(some_dt, max_dt=None):
```
davidmiller, line 120: The word 'clean' in this method name is supposed to imply 'delete' ?

intrahospital_api/deployment_check.py
```diff
@@ -0,0 +1,201 @@
...
+
+                    if max_updated < last_updated:
+                        max_updated = last_updated
+
+    return result, max_updated
+
+
+def get_key(observation_number, lab_test):
+    """
+    A key is the observation number, lab test number and patient id.
```
davidmiller, line 113: Why don't we just use the phrase observation number instead of inventing this 'key' terminology?
Won't that be a confusing layer of indirection in 3 months?

intrahospital_api/deployment_check.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # remove duplicate lab tests
+    lab_tests = list(set(obs_to_lab_test.values()))
+    for lab_test in lab_tests:
+        lab_test.delete()
+
+    return values, max_updated
+
+
+def remove_in_everything(some_list, list_1, list_2):
```
davidmiller, line 135: Is this not just a long way to spell `set(some_list).difference(list_1, list_2)` ?
https://docs.python.org/2/library/stdtypes.html#frozenset.difference

fredkingham:
Move this to the new PR

fredkingham:
This has been moved to be apart of https://github.com/openhealthcare/elcid-rfh/pull/528

PR #551, opened 2018-10-26
removes the unused discharge controller

fredkingham:


fredkingham:
Actually let's create a new pull request to remove the default elcid discharge flow

PR #558, opened 2018-10-29
adds a test for get reference range in the api

fredkingham:
This is just for a coverage bump. The intrahospital_api/api.py could do with more test coverage.


elcid/test/test_api.py
```diff
@@ -326,3 +326,30 @@ def test_patient_found_with_hash(self):
         self.assertEqual(
             response["patient"]["demographics"][0]["first_name"], "Dot"
         )
+
+
+class GetReferenceRangeTestCase(OpalTestCase):
+    def to_obs(self, something):
```
davidmiller, line 332: This method barely saves you any typing but makes the intended API of the method much harder to read...

PR #564, opened 2018-11-15
bumps the labs version to v0.4.1 which excludes lab tests from advanc…

fredkingham:
…ed search


fredkingham:
I'm afraid this does require #565 to run unit tests (everything kinda does!)

Issue #568, opened 2018-11-19
Change Referral reason


fredkingham:
Change Referral reason - Anti TNF to be "immunosepresent"

fredkingham:
look at the branch LTBI-clinic

PR #591, opened 2018-11-20
Immunosuppressant is a superset of Anti-TNF refs #568

fredkingham:


fredkingham:
There's questions about why we aren't using the Opal core model.

The fields are different, here we have:
```
date_of_referral
referral_type
referral_reason
```

In Opal we have
```
internal
referral_organisation
referral_name
date_of_referral
referral_team
referral_type
```

fredkingham:
refs #568

PR #609, opened 2018-11-22
Allows us to record both a TB contact and a previous TB diagnosis 

fredkingham:
screen shot 2018-11-22 at 12 52 09
screen shot 2018-11-22 at 12 52 04


elcid/migrations/0030_auto_20181122_1153.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.8 on 2018-11-22 11:53
+from __future__ import unicode_literals
+
+from django.db import migrations, models
```
davidmiller: Is this migration definitely supposed to be here?

davidmiller:
Can we capitalize the b in Tb for the checkboxes?
And while we're there lowercase the t in (TB) type?

After that I think good to merge

fredkingham:
Great call

screen shot 2018-11-23 at 10 48 36
screen shot 2018-11-23 at 10 48 30

Issue #573, opened 2018-11-19
Add Assessment and a reason for interaction


fredkingham:
Adds an assessment text field and a reason for interaction of "LTBI Assessment" to the inline clinical discussion form.

PR #616, opened 2018-11-23
adds in reason for interaction and assessment into the tb detail page…

fredkingham:
…/patient consultation model refs #573


elcid/migrations/0030_auto_20181123_0849.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
davidmiller: @fredkingham This migration is back... `git add .` again ? ;)

fredkingham: 👍 removed.

PR #621, opened 2018-11-27
Change the method UpstreamLabTest.get_relevant_tests to take a list o…

davidmiller:
…f test name strings rather than be hardcoded on the model.


davidmiller:
@fredkingham is this still relevant ?
(We should either do it or close I think ? )
Perhaps the 500 test case is worth cherry-picking either way - can't remember why I bothered (could just be coverage...

Issue #628, opened 2018-12-10
Add a link to the EDRM


fredkingham:
We then need to send it off for testing from users within the RFH network.

fredkingham:
two pull requests related to this ticket, the first is to the now dead v0.3.7 branch. v0.3.6.2 is the branch this should merge in to.

fredkingham:
This is done and in production

PR #629, opened 2018-12-11
adds in an edrm panel refs #628

fredkingham:
The infection service

screen shot 2018-12-11 at 13 49 05

The TB service
screen shot 2018-12-11 at 14 00 26


davidmiller:
Can we wrap in permissions until we have confirmation it works and try an icon that's not the logo of https://wpforms.com/ ?

PR #633, opened 2018-12-12
renames the field translater to what is now on opal 13

fredkingham:


fredkingham:
This is me going the wrong way

PR #657, opened 2018-12-19
removes the infection related issues panel

fredkingham:


fredkingham:
screen shot 2018-12-19 at 15 25 12

PR #663, opened 2019-01-24
migrations that should have happened when _title got moved to verbose…

fredkingham:
… name

These are migrations that should have happened as a part of https://github.com/openhealthcare/elcid-rfh/commit/863a4b0462aa149d080105baaa379505c3f1055a#diff-cd78c938ec4beaa196abec0c07e0633b


fredkingham:
The other two changes included in the migrations are also hangovers rather than new changes.

I think when the referral route choices were changed 
https://github.com/openhealthcare/elcid-rfh/issues/442 no migration was created.

Also LTBR verbose name looks like it was added after migrations were run
https://github.com/openhealthcare/elcid-rfh/commit/15ce7e396625ce6f6c8024257a3556ee686200b0

PR #674, opened 2019-01-26
Tb v0.3.6.2

fredkingham:
screen shot 2019-01-26 at 14 35 09


davidmiller:
Don't suppose we could do the force push thing that removes the now-merged diffs ?

PR #676, opened 2019-01-28
665 tb symptoms

fredkingham:


fredkingham:
screen shot 2019-01-28 at 23 14 33

fredkingham:
There is a problem that `Fatigue` after you save, does not subsequently show as saved as we have it as a synonym for `Malaise`. Also `Breathlessness` that we have as a synonym for `Shortness of breath` I need to think of a resolution... Easiest would be to switch out and use the main name, however I'm not 100% sure that fatique == malaise..

PR #681, opened 2019-02-02
the lab tests results view vanished in the last release because the t…

fredkingham:
…b view had a hanging div and angular could not take it


fredkingham:
@davidmiller ^^ this just adds a closing </div> that was missing in the tb view

PR #694, opened 2019-02-12
690 moves to opal 13 django 2

fredkingham:
Note the decimal place/float


elcid/api.py
```diff
@@ -1,5 +1,6 @@
 import datetime
 import re
+import decimal
```
davidmiller: We don't need this import ? 
Shouldn't some kind of flake8 check catch that ?

fredkingham: 👍

Issue #700, opened 2019-02-22
Hide spark lines for the timebing


fredkingham:
just show the last 6 results.

fredkingham:
completed and deployed

PR #702, opened 2019-02-22
a light weight change the keeps the sparklines but hides them for the…

fredkingham:
… time being and replaces them with 2 extra rows refs #700


fredkingham:
screen shot 2019-02-25 at 09 01 49

PR #707, opened 2019-03-07
adds notes to the social history record panel, changes the column wid…

fredkingham:
…ths so that all field names are displayed in full, adds colons to match nationality and language

### DIFF
--- a/apps/tb/templates/records/social_history.html
+++ b/apps/tb/templates/records/social_history.html
@@ -1,21 +1,21 @@
 <span ng-show="item.consistency_token">
   <div class="row">
-    <div class="form-group">
-      <label class="control-label col-sm-4">
-        Smoking
+    <div>
+      <label class="control-label col-sm-5">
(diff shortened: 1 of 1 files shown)


apps/tb/templates/records/social_history.html
```diff
@@ -88,15 +88,30 @@
...
-        Mental health issues
+    <div>
+      <label class="control-label col-sm-5">
+        Mental health issues:
       </label>
-      <p class="form-control-static col-sm-8">
-        [[ item.mental_health_issues ]]
+      <p class="form-control-static col-sm-7">
+        <span ng-show="item.mental_health_issues">
+          some history
```
davidmiller: @fredkingham is this debugging markup ?

fredkingham: yep, its a bit ambiguous changed it to yes/no, rebased and pushed

PR #715, opened 2019-03-21
Tb list fix

fredkingham:


elcid/__init__.py
```diff
@@ -80,6 +80,6 @@ def get_menu_items(cls, user):
         if standard_add_patient_menu_item.for_user(user):
             menu_items.append(standard_add_patient_menu_item)
         else:
-            menu_items = [i for i in menu_items if not i.href == "/#/list/"]
```
davidmiller, line 83: I think we need to test these ?

fredkingham:
We don't actually want the TB list visible at the moment

PR #716, opened 2019-03-22
Refactored lab tests

fredkingham:


plugins/labtests/models.py
```diff
@@ -0,0 +1,204 @@
...
+    return dt_format(some_dt, settings.DATETIME_FORMAT)
+
+
+"""
+Models for labtests
+"""
+class LabTest(models.Model):
+    patient = models.ForeignKey(
+        omodels.Patient,
+        on_delete=models.CASCADE,
```
davidmiller, line 20: Does this definitely work the way we want it to ?

plugins/labtests/models.py
```diff
@@ -0,0 +1,204 @@
...
+    status = models.CharField(max_length=256, blank=True, null=True)
+    test_code = models.CharField(max_length=256, blank=True, null=True)
+    test_name = models.CharField(max_length=256, blank=True, null=True)
+    lab_number = models.CharField(max_length=256, blank=True, null=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    @classmethod
+    def create_from_old_test(cls, lt):
+        s = cls()
```
davidmiller: lots of single name variables here...

plugins/labtests/models.py
```diff
@@ -0,0 +1,204 @@
...
+        fields = [
+            "clinical_info",
+            "site",
+            "status",
+            "test_code",
+            "test_name",
+            "lab_number",
+        ]
+        for f in fields:
+            setattr(s, f, extras.get(f, None))
```
davidmiller, line 47: I'd swear a lot if this line ever turned up in a stacktrace :)

plugins/labtests/models.py
```diff
@@ -0,0 +1,204 @@
...
+            "observation_value",
+            "reference_range",
+            "units"
+        ]
+        for field in fields:
+            as_dict[field] = getattr(self, field)
+        return as_dict
+
+
+def create_from_old_tests(patient):
```
davidmiller: I think this whole thing could be lost and make the management command iterate over upstream tests?

elcid/api.py
```diff
@@ -78,6 +80,26 @@
         LAB_TEST_TAGS[test_name].append(tag)
 
 
+def get_upstream_lab_tests_for_patient(user, patient):
```
davidmiller: none of these three functions are tested...

plugins/labtests/models.py
fredkingham: resolved on 9791e5ea

plugins/labtests/models.py
fredkingham: resolved on 83b691d6

plugins/labtests/models.py
fredkingham: resolved on 83b691d6

elcid/api.py
fredkingham: resolved on 781046f6

fredkingham:
@davidmiller I think these are all resolved. Thoughts?

PR #722, opened 2019-04-15
Tb changes for 9 apr

fredkingham:


fredkingham:
- [x] fix tests
- [x] Remove No Travel model field
- [ ] Can we look at the field names on tb history
- [x] check menu items is correct for a TB user
- [x] Remove referral reasons choices (we are now using a lookup list)
- [ ] Add stories to check letters
- [x] The opal change needs unit tests/change log
- [x] Secondary diagnosis has a lower case d
- [x] Adverse reaction needs an icon
- [x] Referral route display sometimes has `:` at the end of the field name
- [ ] Tb Treatment modal, other medication can't click in to end date --> This is generic opal behaviour, first click opens the date modal, second clicks into it.
- [x] Observations can't be editted
- [x] Primary diagnosis is singular
## Follow up letter
- [x] Follow up letter does not show language/country of birth/travel
- [x] TB medication, the date display doesn’t show what the date relates to
- [x] No recorded travel is not showing up
- [x] Can either everything or nothing have periods for the no-entry text.
- [ ] Not sure weight today is working ?  --> I think this works so I'm doing something wrong
- [x] key investigations, make the date 1st and it will align better, lose the parenthesis
## Initial letter
- [ ] not showing country of birth --> Are you sure?
- [x] date ordering for key investigations as ^^
- [ ] Weight as ^^ --> I think this works so I'm doing something wrong
- [x] periods on no-entry text as ^^

elcid/__init__.py
```diff
@@ -80,6 +80,6 @@ def get_menu_items(cls, user):
         if standard_add_patient_menu_item.for_user(user):
             menu_items.append(standard_add_patient_menu_item)
         else:
-            menu_items = [i for i in menu_items if not i.href == "/#/list/"]
```
fredkingham, line 83: I know this was done, so that TB viewers would see the patient lists. I'm not sure why it was left excluding`/`

elcid/models.py
```diff
@@ -637,29 +643,21 @@ class ReferralRoute(omodels.EpisodeSubrecord):
     REFERRAL_REASON = (
         ("Symptomatic", "Symptomatic",),
         ("TB contact screening", "TB contact screening",),
-        ("New entract screening", "New entract screening",),
+        ("New entrant screening", "New entrant screening",),
         ("Transferred in TB Rx", "Transferred in TB Rx",),
-        ("Immunosuppressant", "Immunosuppressant",),
+        ("Pre-immunosuppression screening", "Pre-immunosuppression screening",),
```
fredkingham, line 648: We can remove these referral reasons now right?

fredkingham:
Closing this as it now has been superceded

PR #729, opened 2019-06-02
Remove the old lab tests api

fredkingham:
Actually I'd prefer this to go into v0.12 as it would be nice to leave it running for a week, but I believe it's good to go.


fredkingham:
closing this to repoint to v0.12

PR #731, opened 2019-06-17
V0.12 lab test release

fredkingham:
It depends on https://github.com/openhealthcare/opal/pull/1735

Its this looks like a lot, but its broken down in 2 commits.

1. Remove the old api, just use the new one. This has been running since yesterday and looks good.
2. Swap out the old api, use one that uses the database now, because we can, and add unit tests.


davidmiller:
Why do we have two get x classes defined by a setting ?

fredkingham:
I've got into the pattern of enabling new features by a settings flag. Then removing them when we are happy the new ones work. We could just remove the old.

davidmiller:
```
            If a patient has multiple observations
            on the same date take the most recent
            that can be cast into a number
```

... I'm not so sure this is the right thing to do - how did we come up with it? 
Are you sure it's better than just doing the simple thing e.g. printing the last N results ?

davidmiller:
`enabling new features by a settings flag` right, but this is a bugfix ;)

fredkingham:
> `enabling new features by a settings flag` right, but this is a bugfix ;)
Happy to remove the flag, its just a safety net, that you're right its probably unnecessary.

fredkingham:
> ```
>             If a patient has multiple observations
>             on the same date take the most recent
>             that can be cast into a number
> ```
> 
> ... I'm not so sure this is the right thing to do - how did we come up with it?
> Are you sure it's better than just doing the simple thing e.g. printing the last N results ?

I _think_ this is what was agreed although we think it would be a good idea to reevaluate the whole panel.

From the raw data as we currently save it because we are order, as previously agreed by lab test datetime_ordered, and because upstream insert new rows for pending and then replace them with actual results, we can have multiple results for the exact same datetime_ordered.

There more work to look at whether we can just remove these pending results or how we should handle them. 

This is a bug fix to stop the pending results making the data appear as if there is no result for that date..

elcid/api.py
```diff
@@ -375,107 +355,205 @@ def retrieve(self, request, patient):
...
+                        if date_to_observation[obs_date].observation_datetime < dt:
+                            date_to_observation[obs_date] = i
+                    else:
+                        date_to_observation[obs_date] = i
+                if len(date_to_observation.keys()) == self.NUM_RESULTS:
+                    break
+
+            return date_to_observation
+
+        def get_obs_qs(self, patient, lab_test_name, observation_name):
```
davidmiller: can we use the whole words where possible - e.g. `def get_observation_queryset`

davidmiller:
If what we're doing is coping with an upstream database that behaves in a certain way, then I think the commit should explain that - the code (docstrings, comments) should definitely read that way (there's an argument that the methods etc should be oriented that way) - at the moment the why of the picking the 'last numeric result) is opaque when you read it...

PR #738, opened 2019-07-19
Renal handover

fredkingham:


elcid/templates/episode_list.html
```diff
@@ -31,10 +31,14 @@ <h1 >
                   {% endfor %}
                 </ul>
               </div>
+              <a ng-show="currentTag === 'renal'" href="/letters/#{% url 'renal_handover' %}" target="_blank" class="btn btn-secondary pull-left left-offset-10">
```
davidmiller, line 34: I'm not sure why this has a hard coded href component and a `url` tag ?

elcid/urls.py
```diff
@@ -18,6 +18,11 @@
     url(r'stories/$', views.TemplateView.as_view(template_name='stories.html')),
     url(r'elcid/v0.1/', include(api.elcid_router.urls)),
     url(r'labtest/v0.1/', include(api.lab_test_router.urls)),
+    url(
+        r'^elcid/renal_handover',
```
davidmiller, line 22: We're prefixing urls with the name of the application now ? ;)

elcid/test/test_views.py
```diff
@@ -78,3 +82,270 @@ def setUp(self):
...
+            "ICU 4 East",
+            "ICU 4 West",
+            "Outpatients",
+            "PITU",
+        ]
+        self.assertEqual(
+            expected, sorted(wards, key=views.ward_sort_key)
+        )
+
+    def test_grouping(self):
```
davidmiller, line 118: What is this actually testing ? Can we have a comment / docstring / method name ?

elcid/test/test_views.py
```diff
@@ -78,3 +82,270 @@ def setUp(self):
...
+            "8 West",
+            "8 West",
+            "9 East",
+            "9 North",
+        ]
+        self.assertEqual(
+            expected, sorted(wards, key=views.ward_sort_key)
+        )
+
+    def test_numerical_ordering(self):
```
davidmiller: Same for this one - what specifically does it test that's not the generic case above?

elcid/views.py
```diff
@@ -100,3 +104,105 @@ def get_context_data(self, *args, **kwargs):
...
+
+    start = ward[:2].strip()
+    rest = ward[2:]
+    return (len(start), start, rest)
+
+
+class RenalHandover(LoginRequiredMixin, TemplateView):
+    template_name = "elcid/renal_handover.html"
+
+    def get_unit_ward(self, location):
```
davidmiller, line 135: The terminology is taken from their word document template? 
I think it's odd terminology to replicate in code - get location perhaps... 

Are you sure you only want to display this if they have both bed and ward?

elcid/views.py
```diff
@@ -100,3 +104,105 @@ def get_context_data(self, *args, **kwargs):
...
+            )
+
+    def get_context_data(self, *args, **kwargs):
+        """
+        An ordered dictionary of ward to patient
+
+        patient here is...
+        their name
+        their hospital number
+        their unit/ward
```
davidmiller, line 150: This is actually 'their ward/bed' - stealing the terminology makes the docstring confusing...

elcid/views.py
```diff
@@ -100,3 +104,105 @@ def get_context_data(self, *args, **kwargs):
...
+        that episodes diagnosis
+        all of the clinical advice related to the patients infection service
```
davidmiller, line 153: This hasn't been updated to include new things from the review ?

elcid/views.py
```diff
@@ -100,3 +104,105 @@ def get_context_data(self, *args, **kwargs):
...
+            patient_ids.add(episode.patient_id)
+            location = episode.location_set.all()[0]
+            diagnosis = ", ".join(
+                i.condition for i in episode.diagnosis_set.all()
+            )
+            demographics = models.Demographics.objects.filter(
+                patient__episode=episode
+            ).get()
+            microbiology_inputs = models.MicrobiologyInput.objects.filter(
+                episode__patient__episode=episode
```
davidmiller: Is this definitely right ?
Does it not then put us back in the same episodic advice situation?

elcid/views.py
```diff
@@ -100,3 +104,105 @@ def get_context_data(self, *args, **kwargs):
...
+                "lines": episode.line_set.all(),
+                "clinical_advices": microbiology_inputs,
+                "blood_culture_sets": episode.patient.bloodcultureset_set.all()
+            })
+
+        result = OrderedDict()
+        wards = by_ward.keys()
+        wards = sorted(wards, key=ward_sort_key)
+
+        other = None
```
davidmiller: This 10 lines is 'Group blank wards under "Other" ?

elcid/views.py
```diff
@@ -100,3 +104,105 @@ def get_context_data(self, *args, **kwargs):
...
+        for ward in wards:
+            if not ward:
+                other = by_ward[ward]
+            else:
+                result[ward] = by_ward[ward]
+
+        # add `other` last to the ordered dict
+        if other:
+            result["Other"] = other
+
```
davidmiller: If I'm understanding you right, I think I might have spelled this as below and just not returned an ordered dict ;)
```python
if None in by_ward:
    by_ward["Other"] = by_ward[None]
    del by_ward[None]

ward_names = sorted(by_ward.keys())
episodes_by_ward = [{'ward': w, 'episodes': by_ward[w]} for w in ward_names]
```

elcid/templates/episode_list.html
fredkingham: This is a consequence of the back and forward of the letters plugin, it is a bit weird...

The theory is that we want it rendered within the angular context to give us access to our angular filters e.g. markdown. I'd rather not accidentally have 2 different markdown renderers. Also the copy and paste directive. 

I agree though its a bit awks!

elcid/test/test_views.py
fredkingham: As discussed this is to namespace the views in terms of the service that's using them.
Definitely true that using `elcid` as a sometime synonym for `infection_service` is not idea.

elcid/urls.py
fredkingham: As discussed this is to namespace the views in terms of the service that's using them.
Definitely true that using elcid as a sometime synonym for infection_service is not idea.

elcid/views.py
fredkingham: Great call

elcid/views.py
fredkingham: So we do give me inner join on patient then inner join on the episode table again which is basically expected, but as discussed, let just get the patient and that can save lots of joins.

```
SELECT "elcid_microbiologyinput"."id", "elcid_microbiologyinput"."created", "elcid_microbiologyinput"."updated", "elcid_microbiologyinput"."created_by_id", "elcid_microbiologyinput"."updated_by_id", "elcid_microbiologyinput"."consistency_token", "elcid_microbiologyinput"."episode_id", "elcid_microbiologyinput"."when", "elcid_microbiologyinput"."initials", "elcid_microbiologyinput"."infection_control", "elcid_microbiologyinput"."clinical_discussion", "elcid_microbiologyinput"."agreed_plan", "elcid_microbiologyinput"."discussed_with", "elcid_microbiologyinput"."clinical_advice_given", "elcid_microbiologyinput"."infection_control_advice_given", "elcid_microbiologyinput"."change_in_antibiotic_prescription", "elcid_microbiologyinput"."referred_to_opat", "elcid_microbiologyinput"."white_cell_count", "elcid_microbiologyinput"."c_reactive_protein", "elcid_microbiologyinput"."maximum_temperature", "elcid_microbiologyinput"."liver_function_fk_id", "elcid_microbiologyinput"."liver_function_ft", "elcid_microbiologyinput"."reason_for_interaction_fk_id", "elcid_microbiologyinput"."reason_for_interaction_ft", "elcid_microbiologyinput"."renal_function_fk_id", "elcid_microbiologyinput"."renal_function_ft" FROM 
"elcid_microbiologyinput" INNER JOIN "opal_episode" ON ("elcid_microbiologyinput"."episode_id" = "opal_episode"."id") INNER JOIN "opal_patient" ON ("opal_episode"."patient_id" = "opal_patient"."id") INNER JOIN "opal_episode" T4 ON ("opal_patient"."id" = T4."patient_id") WHERE T4."id" = 12 ORDER BY "elcid_microbiologyinput"."when" DESC; args=(12,)
```

elcid/test/test_views.py
fredkingham: good call, removed

elcid/views.py
fredkingham: Changed so it will just show the ward if we are missing a bed

elcid/views.py
fredkingham: updated

elcid/views.py
fredkingham: we actually have episode.patient_id so we can remove the join with no actual cost. updated.

elcid/views.py
fredkingham: implemented a version of this

PR #740, opened 2019-07-19
Adds the add slashes to the markdown in the ltbi letters this is beca…

fredkingham:
…use people use quotation marks in their clinical advice that breaks when rendered raw


fredkingham:
Closing to reopen on the latest branch

PR #756, opened 2019-08-29
Adds the plural filter which is no longer in opal

fredkingham:


fredkingham:
This was undone and changed to the v0.16 branch

PR #761, opened 2019-10-11
Reduce some of the padding on the test result view and add a 7th obse…

davidmiller:
…rvation where available.


fredkingham:
👍

fredkingham:
This was undone and changed to the v0.16 branch

PR #760, opened 2019-10-11
Dont' reverse observation dates in the test results detail view

davidmiller:

PR #763, opened 2019-10-25
Revert "Dont' reverse observation dates in the test results detail view"

fredkingham:
Reverts openhealthcare/elcid-rfh#760 as it should go into v0.16


fredkingham:
This reversion was reverted

PR #765, opened 2019-11-22
This changes the lab tests api so that we only pull 7 most recent res…

fredkingham:
…ults accross for each observation


fredkingham:
This changes it so that texts in long form (ie not a series of numbers in a table but a list of text results) which were previously not limitted to 7 results, will now only show 7 results.
Time wise on a complicated patient this brings the time down from 5.5 seconds to 5.1 seconds
not a lot but we expect the real improvement when we move more logic into the db.

PR #767, opened 2019-11-30
Lab test api refactor

fredkingham:
So this speeds lab test loads up by ~3x (a bit faster)

It also deletes a lot of unused code.


davidmiller:
`aggregate_observations_by_lab_test` is quite an odd method... 

it's like a 6 line aggregation method that has a 30 line `serialise_observation` method in the middle of it :)

I would definitely be tempted to make those separate methods TBH

davidmiller:
Particularly because we then do a whole other pass of 'serialise observation' stuff in the 'retrieve' method I think ? :)

PR #772, opened 2019-12-19
Lab test cleanup 2

fredkingham:
So this means we use the same method for both the lab test summary api end point and the results view end point. 

It also fixes the bug where observations of a form such as 22.08.2019 cause the lab tests to fail

### DIFF
--- a/elcid/api.py
+++ b/elcid/api.py
@@ -201,6 +201,33 @@ class LabTestResultsView(LoginRequiredViewset):
             )
         return observation_metadata
 
+    def sort_lab_tests_dicts(self, lab_test_dicts):
+        """
+        Ordered by most recent observations first please, this means
+        the first date if its a long form observation or the last
+        if its not. (as short form dates are shown ascending and long form descending)
+
+        When there are multiple results for the same date, tests should be shown in
+        alphabetical order.
+        """
+        serialised_tests = sorted(
+            lab_test_dicts, key=itemgetter("lab_test_type")
+        )
+
+        def get_latest_date(obs_dict):
+            if obs_dict["long_form"]:
+                return obs_dict["observation_date_range"][0]
+            else:
+                return obs_dict["observation_date_range"][-1]
+
+        # ordered by most recent observations first please, bu maintaining
+        # the alphabetically order of the lab tests
+        serialised_tests = sorted(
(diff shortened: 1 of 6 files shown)


elcid/api.py
```diff
@@ -179,15 +129,84 @@ def is_empty_value(self, observation_value):
         else:
             return observation_value is None
 
+    def get_qs(self, patient):
```
davidmiller: `get_last_365_days` perhaps ?

elcid/api.py
```diff
@@ -179,15 +129,84 @@ def is_empty_value(self, observation_value):
...
+        lab_tests = lab_tests.exclude(
+            test_name__in=to_ignore
+        )
+        lab_tests = lab_tests.filter(datetime_ordered__gte=a_year_ago)
+        return lab_tests.prefetch_related('observation_set')
+
+    def get_observations_by_type_and_date_str(self, observations):
+        """
+        Returns a dict of observation name -> date -> observation
+        We choose the most recent date for the datetime unless the most
```
davidmiller: Are Date and datetime the wrong way round in this sentence?

elcid/api.py
```diff
@@ -179,15 +129,84 @@ def is_empty_value(self, observation_value):
...
+        recent is 'Pending'
+        """
+        result = defaultdict(dict)
+        observations = sorted(
+            observations, key=lambda x: x.observation_datetime
+        )
+        observations.reverse()
```
davidmiller: Doesn't providing your own sorting function only to reverse it on the next line seem odd?
`- ....to_timestamp()` or similar ?

elcid/api.py
fredkingham: 👍

elcid/api.py
fredkingham: So we were doing 

```
        observations = sorted(
            observations, key=lambda x:- x.observation_datetime.toordinal()
        )
```

I think this is the new way is more readable. I think I prefer this way but I'm happy with either.

elcid/api.py
fredkingham: 👍

elcid/api.py
fredkingham: much better idea... use the `reverse=True` argument...

PR #786, opened 2020-01-20
Removing old lab tests

fredkingham:
So this is a beginner PR to remove the apparatus around the old lab tests.

It deletes all the old
* blood culture pathways
* api end points
* management commands for synching things.


fredkingham:
I'm not sure this should go into v0.19 as that will probably be the upstream database change. Probs the one after!

PR #791, opened 2020-01-24
Changes update logic to delete all existinb observations as this is w…

fredkingham:
…hat happens in the table


fredkingham:
unnecessary after https://github.com/openhealthcare/elcid-rfh/pull/794

PR #823, opened 2020-02-13
Antifungal add patients

fredkingham:
This doesn't do any tag manipulation it just adds a patient and loads in tests if they don't already exist.

Tested with a sample version of the antifungal spread sheet


elcid/models.py
```diff
@@ -344,6 +344,17 @@ class Antimicrobial(EpisodeSubrecord):
     _icon = 'fa fa-flask'
     _modal = 'lg'
 
+    ANTIFUNGAL_DRUGS = [
```
davidmiller: I don't think this is still used?

elcid/models.py
```diff
@@ -395,6 +408,17 @@ class Meta:
         verbose_name = "Clinical Advice"
         verbose_name_plural = "Clinical Advice"
 
+# method for updating
+@receiver(post_save, sender=MicrobiologyInput)
+def update_chronic_antifungal_reason_for_interaction(
+    sender, instance, **kwargs
+):
+    asr = MicrobiologyInput.ANTIFUNGAL_STEWARDSHIP_ROUND
```
davidmiller, line 405: This is about line length ?

elcid/models.py
```diff
@@ -758,6 +782,47 @@ def get_api_name(cls):
         return camelcase_to_underscore(cls._meta.object_name)
 
 
+class ChronicAntifungal(omodels.PatientSubrecord):
```
davidmiller: Is this definitely a patientsubrecord? e.g. should it also have it's own API endopint, and be serialised on every episode/patient context?

elcid/models.py
```diff
@@ -758,6 +782,47 @@ def get_api_name(cls):
...
+    DRUG_LIST = "drug_list"
```
davidmiller: I think drug list is ambiguous here - first pass I read this as 'reason == something to do with a list of drugs' whereas it's the dispensary report rather than something like the list of `    ANTIFUNGAL_DRUGS = [ ` above?

elcid/models.py
```diff
@@ -758,6 +782,47 @@ def get_api_name(cls):
...
+    )
+    updated_dt = models.DateTimeField(auto_now=True)
+    reason = models.TextField(
+        choices=REASONS, blank=True, null=True
+    )
+
+    ACTIVE_DAYS = 3
+
+    @classmethod
+    def active_from_date(cls):
```
davidmiller: This, in conjunction with active being a property is kind of confusing - you mean something like 'return me the date at which which we consider things that have been updated to be either active or inactive' ?

elcid/models.py
```diff
@@ -758,6 +782,47 @@ def get_api_name(cls):
...
+        return timezone.now().date() - datetime.timedelta(
+            cls.ACTIVE_DAYS
+        )
+
+    @property
+    def active(self):
+        active_from_date = self.__class__.active_from_date()
+        days_ago = datetime.date.today() - datetime.timedelta(
```
davidmiller: Is this right ? Why would this be today and then compared to a classmethod that looks at now() ?

fredkingham:
@davidmiller I think its all fixed.

elcid/assets/js/elcid/controllers/add_antifungal_patients.js
```diff
@@ -0,0 +1,144 @@
...
+   var newPatient = _.clone(patient);
+   newPatient.demographics = {hospital_number: hospitalNumber};
+   $scope.lookup(newPatient);
+   return newPatient;
+  }
+
+  $scope.getSearchArgs = function(patientForm){
+    var searchArgs = {}
+    searchArgs[DemographicsSearch.PATIENT_FOUND_IN_ELCID] = function(result){
+      patientForm.demographics = result.demographics[0];
```
davidmiller: Why do we write this out twice? it does the same thing no?

elcid/assets/js/elcid/controllers/add_antifungal_patients.js
```diff
@@ -0,0 +1,144 @@
...
+    if(!patientForm.demographics.hospital_number || !patientForm.demographics.hospital_number.length){
+      return
+    }
+    patientForm.error = undefined;
+    patientForm.state = $scope.states.SEARCHING;
+    var args = $scope.getSearchArgs(patientForm);
+    DemographicsSearch.find(patientForm.demographics.hospital_number, args);
+  }
+
+  $scope.research = function(){
```
davidmiller: Can we rename to e.g. `re_search` to avoid it being read as "research() and development()`

elcid/patient_lists.py
```diff
@@ -153,25 +157,53 @@ class Bacteraemia(RfhPatientList):
...
+
+class ChronicAntifungal(RfhPatientList, PatientList):
+    display_name = "Chronic Antifungal"
+    template_name = 'episode_list.html'
+    schema = []
+    is_chronic = True
+
+    @property
+    def queryset(self):
+        active_from_date = datetime.date.today() - datetime.timedelta(3)
```
davidmiller, line 194: I'd be tempted to avoid the second query and re-sorting e.g. 

```
 def get_queryset(self, user):
     active_from_date = datetime.date.today() - datetime.timedelta(3)
     episodes = omodels.Episode.objects.filter(
         patient__chronicantifungal__updated_dt__gte=active_from_date
     ).filter(
         category_name=InfectionService.display_name
     ).order_by('created')

     patients_seen = set()
     serialise = []

     for episode in episodes:
         if episode.patient_id in patients_seen:
             continue
         serialise.append(episode)
         patients_seen.add(episode.patient_id)

    return serialise
```

elcid/templates/partials/add_patients/lookup_patients.html
```diff
@@ -0,0 +1,22 @@
+<div class="form-group">
+  <form ng-submit="lookupAll()" {% if error %}class="has-error"{% endif %} method="POST">
+    <h3>
+      Hospital numbers
+    </h3>
+    <p class="lead">
+      Please enter a list of all hospital numbers for the patients you would like to add to elcid
```
davidmiller: I don't think we capitalise it like that ?

elcid/patient_lists.py
```diff
@@ -153,25 +157,53 @@ class Bacteraemia(RfhPatientList):
...
     tag = "liver_transplantation"
     template_name = 'episode_list.html'
     schema = []
+
+
+class ChronicAntifungal(RfhPatientList, PatientList):
+    display_name = "Chronic Antifungal"
+    template_name = 'episode_list.html'
+    schema = []
+    is_chronic = True
```
davidmiller: Do we want something like 'display_footer_buttons' etc ? 

It's not going to be just chronic as part of this release...

PR #833, opened 2020-02-21
Adds a risk factor panel to elcid

fredkingham:
Things to note.

Its called "Risk Factors Present" because that's the name used on the CQUIN.

We are not recording start and end, mostly because in general we don't and it would add additional fields to fill in. Thoughts?


fredkingham:
superseded by https://github.com/openhealthcare/elcid-rfh/pull/837

PR #834, opened 2020-02-21
renames antimicrobials to anti-infectives. We don't change the model …

fredkingham:
…name as we are thinking about changing/merging this model into treatments


davidmiller:
Wouldn't this generate a migration?

PR #837, opened 2020-02-21
Adds a risk factor panel to elcid

fredkingham:


elcid/models.py
```diff
@@ -707,6 +707,18 @@ def __str__(self):
         return self.name
 
 
+class PatientRiskFactor(lookuplists.LookupList):
+    pass
+
+
+class RiskFactor(omodels.PatientSubrecord):
+    risk_factor = ForeignKeyOrFreeText(PatientRiskFactor)
+    _icon = 'fa fa-map-signs'
```
davidmiller: We have these properties above the fields by unspoken convention I think... 
And, I would have picked the warning triangle with exclamation mark - what does map signs look like?

PR #846, opened 2020-02-28
minor change to allow the api url to be passed through to the test su…

fredkingham:
…mmary panel


fredkingham:
This is going in as part of a bigger PR

PR #850, opened 2020-03-02
We are not using lab tests in the letters at the moment, so stop addi…

fredkingham:
…ng them to the context


davidmiller:
If I merge this will you pull it into the antifungal branch too ?

PR #859, opened 2020-03-24
No need for an HR when there's nothing below it

davidmiller:


fredkingham:
I think we're closing one more </div> than we should be.

fredkingham:
> I think we're closing one more than we should be.

Actually looks like we were already closing one more than we should have been.

PR #862, opened 2020-03-24
Don't show the seconds in the result view

fredkingham:
don't show `24/04/2020 09:25:32`
show `24/04/2020 09:25`


davidmiller:
You definitely want to do this with substring in a template on the client, not in the `    def serialise_long_form_instance(self, instance):' ?

PR #864, opened 2020-03-24
Generates a secret key on deployment and puts it into the local setti…

fredkingham:
…ngs file

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -44,6 +44,7 @@ import datetime
 import json
 import copy
 import time
+import random
 from jinja2 import Environment, FileSystemLoader
 
 from fabric.api import local, env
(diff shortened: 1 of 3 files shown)


fabfile.py
```diff
@@ -326,12 +327,18 @@ def services_symlink_upstart(new_env):
         symlink_name
     ))
 
+def generate_secret_key():
+    chars = 'abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)'
+    return ''.join([
+        random.SystemRandom().choice(chars) for i in range(50)
+    ])
```
davidmiller, line 334: Can't you https://github.com/django/django/blob/3c447b108ac70757001171f7a4791f493880bf5b/django/core/management/utils.py#L82 ?

Issue #855, opened 2020-03-23
ICU Consult Proforma


davidmiller:
For Clinical Advice, when reason for interaction is ICU round, some additional fields to enter and display:

Ventilation
 - Ventilated [Y | N]
 - Type: [NIV | Intubated]
 - Fio2: [char]
 - Comments [Text]

Inotropic Support
 - Inotropic [Y | N]
 - Drug [Char lookuplist]
 - Dose [Char]

Others
- Meld Score 
- Sofa Score
- Temperature 

Model suggestion:

```python

Class ICUStatus:
  date = DateField()
  ventilation_type = ChoicesField()
  fi02 = CharField()
  ventilation_comments = TextField()
  inotropic_drug = LookuplistField()
  inotropic_dose = CharField()
  meld_score = CharField()
  sofa_score = CharField()
```

Inotropic drugs: 
noradrenaline
adrenaline
dobutamine
vasopressin
inoximone
milrinone
levosemendin
metaraminol

blu3id:
To be a pedant:
There is a significant difference between inotropes and vasopressors. Some drugs do both (adrenaline) others have a predominant focus.

To save complexity of trying to correctly classify each would suggest some form of naming for such drugs as "cardiovascular support"

davidmiller:
Thanks for being a pedant @blu3id  :) 

In the end we wound up separating Vasopressor and Inotropes - this is now reviewed and pending a deployment this evening.

PR #877, opened 2020-03-28
Icu round

fredkingham:
Tis a tale of three parts.
1. Refactors a general edit controller that fits with all the hooks for our modal template and handles deletion as long as you pass in an object that has a save method and a delete method and a call back for what you want done afterwards. Then changes the blood culture isolate to use it (it pretty much did already but now its called general edit rather than blood culture...)

2. The back end models we save in the format of.
```
microinput: {
  micro_input_icu_round_relation: {
      observation: { 
          "temperature": 100
      },
      icu_round: {
           "fio2": 20
           ...
      }
  }
  clinical_advice: "..."
  [[ other fields }}
}
```
 If you delete and observation or an icu round, it does not cascade but if you delete a microinput it will cascade and delete the attatched. Similarly if you change the reason for interaction away from "ICU round" it will delete the Micro ICU Round Relation, observation and ICU Round.

3. Javascript "ClinicalAdvice.js" which wires in to the generic template, its a wrapper for item.js that allows us to use the nested relationship. It also makes sure checkboxes when using inline and modal forms together (ie it gives them a unique name field).


fredkingham:
Screen Shot 2020-03-28 at 14 51 56

Screen Shot 2020-03-28 at 14 51 08

Screen Shot 2020-03-28 at 14 50 58

Screen Shot 2020-03-28 at 14 51 36

fredkingham:
ticket https://github.com/openhealthcare/elcid-rfh/issues/855

PR #882, opened 2020-04-03
Covid dashboard

davidmiller:


plugins/covid/management/commands/calculate_covid_dashboard.py
```diff
@@ -0,0 +1,38 @@
...
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        CovidDashboard.objects.all().delete()
+        print('Deleted Dashboard objects')
+
+        tested = Patient.objects.filter(
+            lab_tests__test_name=CORONAVIRUS_TEST_NAME).distinct()
+        print('Found Tested Patients')
```
fredkingham: Not actually true, 
"Queried for tested patients" is more accurate

plugins/covid/management/commands/calculate_covid_dashboard.py
```diff
@@ -0,0 +1,38 @@
...
+        print('Found Positive Patients')
+
+        negative = tested.filter(
+            lab_tests__observation__observation_value__in=NEGATIVE_TEST_VALUES
+        ).exclude(id__in=positive_ids)
+
+        CovidDashboard.objects.create(
+            patients_tested=tested.count(),
+            positive=positive.count(),
+            negative=negative.count()
```
fredkingham, line 37: We expect
patients_tested == positive + negative?

(is it worth sending an email if they don't?)

plugins/covid/management/commands/calculate_covid_dashboard.py
```diff
@@ -0,0 +1,38 @@
...
+
+        tested = Patient.objects.filter(
+            lab_tests__test_name=CORONAVIRUS_TEST_NAME).distinct()
+        print('Found Tested Patients')
+
+        positive = tested.filter(
+            lab_tests__observation__observation_value__in=POSITIVE_TEST_VALUES
+        )
+        positive_ids = [p.id for p in positive]
+        print('Found Positive Patients')
```
fredkingham: Does the distinct() clause definitely carry over to here? I think it _should_ but I'm not 100%

IE are we sure that if a patient has multiple positive test values they won't be duplicated here?

plugins/covid/models.py
```diff
@@ -0,0 +1,15 @@
...
+from django.db import models
+
+
+class CovidDashboard(models.Model):
+    """
+    Stores figures for the Covid 19 Dashboard in the
+    database when they take a significant amount of time
+    to calculate.
+    """
+    patients_tested = models.IntegerField()
```
fredkingham, line 13: I think we should add auto add fields (ie created timestamps)

plugins/covid/plugin.py
```diff
@@ -0,0 +1,33 @@
...
+
+        dashboard = menus.MenuItem(
+            href='/#/covid-19/',
+            display='Covid 19',
+            icon='fa fa-dashboard',
+            activepattern='/#/covid-19/'
+        )
+
+        profile = UserProfile.objects.get(user=user)
+        if profile.roles.filter(name=COVID_ROLE).exists():
```
fredkingham, line 27: if I'm `user.profile.roles.filter(name=COVID_ROLE).exist()` looks a bit nicer but I that's just my aesthetic opinion (very unimportant)

plugins/covid/management/commands/calculate_covid_dashboard.py
davidmiller: yes

plugins/covid/models.py
davidmiller: Not sure there's any point. Created is always arbitrarily 12 minutes past the last hour b/c the cron job and the delete all. We only need one row in this table, we're not storing history.

PR #887, opened 2020-04-07
Split Cardiovascular support into two.

davidmiller:


elcid/templates/records/icu_round.html
```diff
@@ -14,10 +10,22 @@
...
+<span ng-show="item.inotrope_dose">
     [[ item.inotropic_dose ]]
    <br />
 </span>
+<span ng-show="item.inotrop">
+    [[ item.inotropic_drug ]]
+   <br />
+</span>
+<span ng-show="item.vasopressor_dose">
+    [[ item.inotropic_dose ]]
```
fredkingham: I think this should be `vasopressor_dose`

elcid/models.py
```diff
@@ -841,17 +845,17 @@ class ICURound(EpisodeSubrecord):
...
         max_length=200, blank=True, null=True, choices=VENTILATION_TYPES
     )
     fio2 = models.CharField(
         max_length=20,
         blank=True, null=True, verbose_name="FiO₂"
     )
-    inotropic = models.NullBooleanField(blank=True)
-    inotropic_drug = ForeignKeyOrFreeText(InotropicDrug)
-    inotropic_dose = models.CharField(max_length=200, blank=True, null=True)
+    inotrope         = ForeignKeyOrFreeText(InotropicDrug)
```
fredkingham, line 855: Do you want to rename `InotropicDrug`?

davidmiller: not particularly - do you care lots ?

elcid/templates/records/icu_round.html
davidmiller: Sure - not even the only typo in this file.
It's never rendered, so reasonably meh.

PR #905, opened 2020-04-17
Add the model ICULocation

davidmiller:
We are going to add patient lists that build on ward as defined by this location model to see if it's accurate enough that you don't need to do the manual gardening of patients on and off lists.


fredkingham:
@davidmiller can we get a unit test for parse_icu_location looks like a thing that's fairly easy to check, but looking at it, I'm also not 100% sure the data comes in looks like.

PR #908, opened 2020-04-18
Appointments 

davidmiller:
This is implementation stage 1.

This does not add the cron job that runs the sync periodically - the initial import takes many hours, and although the second run should be much faster, suitable to 1/2 times per day, we may need an alternative strategy.

This adds a detail view if and only if the patient has appointments.

Tests are a known work in progress, pending finalising the import strategy.

This should be mergeable as is - there will be no user visible changes until the management command fetch_appointments is run.

The deployment plan would be one run that does this, and manually running the command once, waiting till it finishes, then a second release that contains a periodic sync command and cron job.

The third stage will be to use appointment booking as a way to
a) include historic TB patients in the elCID cohort
b) automatically register and load patients when they have a TB appointment booked


elcid/urls.py
```diff
@@ -10,6 +10,7 @@
 from elcid import api
 
 from elcid import views
```
fredkingham, line 12: What's this doing?

davidmiller: It's just from the feb version  when we had the TB cliniclist view in urls.

plugins/appointments/loader.py
```diff
@@ -0,0 +1,119 @@
...
+
+Q_GET_ALL_PATIENT_APPOINTMENTS = """
+SELECT *
+FROM VIEW_ElCid_CRS_OUTPATIENTS
+WHERE vPatient_Number = @mrn
+AND insert_date > @insert_date
+"""
+
+
+def non_date_identical(frist, second):
```
fredkingham, line 23: frist the worst, second the best?

fredkingham: Also `non_date_identical` meaning `identical_except_date_fields`?

plugins/appointments/loader.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    Given a dictionary of APPOINTMENT data from the upstream
+    database, and the PATIENT for whom it concerns, decide
+    whether to save or discard it, then do so.
+    """
+    appointment_id = appointment['Appointment_ID']
+    # Find out if this appointment has been seen before
+    if Appointment.objects.filter(appointment_id=appointment_id).exists():
+
+        frist = Appointment.objects.get(appointment_id=appointment_id)
```
fredkingham, line 49: more frist

plugins/appointments/management/commands/initial_tb_import.py
```diff
@@ -0,0 +1,70 @@
...
+        results = self.api.results_for_hospital_number(mrn)
+        update_lab_tests.update_tests(patient, results)
+        print('Done')
+        return patient
+
+    def handle(self, *args, **kwargs):
+        """
+        Main entrypoint for loading all TB appointments ever.
+        """
+        raise NotImplementedError('This command has not been refactored - do not run!')
```
fredkingham: do we definitely want this?

plugins/appointments/templates/appointments/clinic_list.html
```diff
@@ -0,0 +1,33 @@
...
+{% block content %}
+  <div class="row content-offset">
+    <div class="col-md-10 col-md-offset-1">
+      <h2 >Upcoming TB Clinic Appointments</h2>
+    </div>
+  </div>
+  {% for appointment in object_list %}
+    <div class="row"p>
+      <div class="col-md-2 col-md-offset-1">
+        {{ appointment.start_datetime }}
```
fredkingham: Wil this definitely format the datetime in the way we want?

plugins/appointments/tests/test_loader.py
```diff
@@ -0,0 +1,75 @@
...
+        }
+        dict_two = {
+            'Appointment_ID'            : '1234',
+            'Appointment_Start_Datetime': '2019-01-01 12:00:00'
+        }
+        appointment = Appointment(**dict_one)
+        self.assertFalse(loader.non_date_identical(appointment, dict_two))
+
+
+class SaveOrDiscardTestCase(OpalTestCase):
```
fredkingham, line 53: this seems less than useful

plugins/appointments/tests/test_loader.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def test_save_appointment(self):
+        appointment_data = {
+
+        }
+        pass
+
+
+class LoadAppointmentTestCase(OpalTestCase):
+
```
fredkingham, line 67: this seems less than useful

PR #924, opened 2020-04-21
Adds in the password reset package and overrides the login template t…

fredkingham:
…o include it


davidmiller:
- [ ] Domain name in the email isn't set up / coming through correctly
- [ ] We need an upstream version to pin to

PR #927, opened 2020-04-21
Refactor ICU

davidmiller:
Move the logic out of a management command into `icu.loader` 

Add unittests

### DIFF
--- a/plugins/icu/loader.py
+++ b/plugins/icu/loader.py
@@ -12,18 +12,27 @@ from plugins.icu import logger
 from plugins.icu.models import ICUHandover, ICUHandoverLocation, parse_icu_location
 from plugins.icu.episode_categories import ICUHandoverEpisode
 
+
 Q_GET_ICU_HANDOVER = """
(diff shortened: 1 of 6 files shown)


plugins/icu/loader.py
```diff
@@ -0,0 +1,73 @@
...
+        our_handover, _ = ICUHandover.objects.get_or_create(patient=patient)
+
+        for k, v in result.items():
+            setattr(
+                our_handover,
+                ICUHandover.UPSTREAM_FIELDS_TO_MODEL_FIELDS[k],
+                v
+            )
+
+        our_handover.save()
```
fredkingham, line 61: I think you mentioned this before we should set the created/updated timestamp (not blocking, I know its a refactor but now might be a good time)

davidmiller: In relation to this ? 
I don't think so.
There's another ticket that does 'alter behaviour of the ICU integration'.
This is explicitly identical behaviour, different code structure, with tests, nothing else.

PR #949, opened 2020-05-13
Covid followup

davidmiller:


plugins/covid/templates/records/covid_admission.html
```diff
@@ -0,0 +1,245 @@
...
+  </div>
+  <div class="col-md-3">
+    [[ item.respiratory_rate ]] Breaths/min
+  </div>
+
+  <div class="col-md-3">
+    <b>Heart Rate:</b>
+  </div>
+  <div class="col-md-3">
+    [[ item.heart_rate ]] Breaths/min
```
fredkingham: Is this right?

plugins/covid/models.py
```diff
@@ -42,3 +58,384 @@ class CovidPatient(models.Model):
...
+class CovidFollowUpCall(EpisodeSubrecord):
+    """
+    A phone call to a patient seeking follow up on their COVID-19 admission
+    """
+    _icon = 'fa fa-phone'
+
+    ONE_TO_TEN           = enum('1', '2', '3', '4', '5', '6', '7', '8', '9', '10')
+    ONE_TO_NINE          = enum('1', '2', '3', '4', '5', '6', '7', '8', '9')
+    ONE_TO_FIVE          = enum('1', '2', '3', '4', '5')
+    ZERO_TO_THREE        = (
```
fredkingham, line 359: does enums with different fields definitely do what we think it does?

plugins/covid/templates/forms/itu_admission_form.html
```diff
@@ -0,0 +1,5 @@
+{% load forms %}
```
fredkingham, line 1: this form was created by scaffold but we don't want it?

plugins/covid/models.py
davidmiller: ? What do you mean ? What do you think it does?

PR #955, opened 2020-05-19
Migrate lookup lists from opal to elcid

fredkingham:


fredkingham:
Note we are not using the model in lab that I deleted.

elcid/migrations/0048_auto_20200518_1456.py
```diff
@@ -0,0 +1,66 @@
+# Generated by Django 2.0.13 on 2020-05-18 14:56
+
+from django.db import migrations
+
+
+def forwards(apps, schema_editor):
+    """
+    Create models for all Microbiology_organism
```
davidmiller: You mean lookuplist entries rather than models?

PR #957, opened 2020-05-22
Get_batch_start_time docstring

davidmiller:


intrahospital_api/loader.py
```diff
@@ -250,25 +250,22 @@ def batch_load(force=False):
...
-        but we have extra cracks to paper over...
+    however...
 
-        a patient load starts during the batch A load, at 9:58. It finishes
-        the db load from the upstream db at 9:59 but is still saving the data
-        to our db at 10:00
-
-        it is excluded from the batch A, so batch B goes and starts its run
-        from 9:59 accordingly.
+    A patient load starts during the batch A load, at 9:58. It finishes
```
davidmiller, line 263: This and line 256 seem to be in conflict ?

fredkingham: I don't _think_ they conflict, however it is not what the code does, which is potentially worse.

It should state that the batch load starts at 9:58 accordingly.

davidmiller: "Batch A starts at 10:00"
"starts during batch A, at 9:58 "

Surely these can't both be accurate...

fredkingham: ah thats fair, its _supposed_ to mean, patient load runs during batch load A.

It starts at 9:58 and runs for e.g. 3 mins, batch load A starts at 10, so the patient load has been running for 2 mins when Batch Load A starts

PR #960, opened 2020-05-22
Admissions

davidmiller:


plugins/admissions/models.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    # We link the upstream data to our system via a single foreign key to the
+    # Patient object. In practice we take the MRN fields and match them
+    # with patient.demographics().hospital_number
+    patient = models.ForeignKey(Patient, on_delete=models.CASCADE, related_name='encounters')
+
+    # Note: Field - level comments here are from the documentation provided by the integration team
+    upstream_id                        = models.IntegerField(blank=True, null=True)              # Internal SQL Database identifier - Primary Key
+    msh_7                              = models.DateTimeField(blank=True, null=True)             # HL7 Message Datetime sent from Cerner
+    msh_9_msg_type                     = models.CharField(blank=True, null=True, max_length=4)   # Last HL7 Message update Type
```
fredkingham, line 37: under the covers postgres stores the chars field in a text field anyway so including max_length can only burn us and not provide any benefit (not a blocker, but maybe we should just use text fields in future)

plugins/admissions/static/js/admissions/admission_view_controller.js
```diff
@@ -0,0 +1,31 @@
+angular.module('opal.controllers').controller('AdmissionView', function(
+    $scope, $http, $q, $window, ngProgressLite
+){
+    "use strict";
+    var vm = this;
```
fredkingham, line 5: we could check here if patient encounter status is true? Your call

PR #962, opened 2020-05-26
When you login redirect to the angular page that you were on

fredkingham:


elcid/templates/registration/login.html
```diff
@@ -22,7 +22,7 @@ <h1>
           <p>Your username and password didn't match. Please try again.</p>
           {% endif %}
 
-          <form method="post" action="{% url 'login' %}?next=/" class="form-horizontal">
+          <form method="post" action="" class="form-horizontal">
```
davidmiller, line 25: Does this mean that if we do somehow get a login box in a modal that it will post to the wrong url ?

fredkingham: If you get a logout box in a modal and click save it redirects to the login page, ie it works fine, still not the ideal which would be to be redirected to the login instead of pulling up a modal.

PR #965, opened 2020-05-28
ICU occupancy dashboard

davidmiller:


plugins/icu/migrations/0006_auto_20200528_1356.py
```diff
@@ -0,0 +1,13 @@
+# Generated by Django 2.0.13 on 2020-05-28 13:56
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
```
fredkingham, line 8: Do we need this?

PR #966, opened 2020-06-01
Covid test codes

davidmiller:
"Covid Test Positive" is a term that describes any possible positive test from what we view as 5 different lab tests. Our Covid lab test abstraction was only used to calculate the Covid dashboard, and only included 3 of these.

Use the abstraction layer that wraps all tests in all 3 places we should be:
- Pre-loading covid cohort patients
- Calculating statistics for the dashboard
- Displaying the covid results ticker

Update to include all relevant tests.


plugins/covid/lab.py
```diff
@@ -145,3 +213,48 @@ def positive(test):
     covid_test = _get_covid_test(test)
     result = get_result(test)
     return result in covid_test.POSITIVE_RESULTS
+
+
+def get_covid_result_ticker(patient):
+    """
+    Given a PATIENT, return a list of dictionaries representing the last
+    3 results for each test type, with a timestamp, result string, and test name.
```
fredkingham, line 221: it looks like the last 2 results? Or am I reading it wrong

plugins/covid/lab.py
```diff
@@ -145,3 +213,48 @@ def positive(test):
...
+    for test in COVID_19_TESTS:
+        data = []
+
+        tests = lab_test_models.LabTest.objects.filter(
+            test_name=test.TEST_NAME, patient=patient
+        ).order_by('-datetime_ordered')
+
+        for test in tests:
+            if len(data) > 2:
+                continue
```
fredkingham, line 234: break rather than continue is the behaviour we want? (Although no real difference)

plugins/covid/loader.py
```diff
@@ -0,0 +1,38 @@
...
+from intrahospital_api.loader import create_rfh_patient_from_hospital_number
+
+from plugins.covid import lab, loader
+
+Q_GET_COVID_IDS = """
+SELECT DISTINCT Patient_Number FROM tQuest.Pathology_Result_view
+WHERE OBR_exam_code_Text = @test_name
+"""
+
+def pre_load_covid_patients():
```
fredkingham, line 16: When the db is being updated this query will error out 

https://github.com/openhealthcare/elcid-rfh/blob/v0.39/intrahospital_api/apis/prod_api.py#L78

(I'm not sure if this is still an issue since they did their db update but it probably is)

plugins/covid/lab.py
```diff
@@ -145,3 +213,48 @@ def positive(test):
...
+                {
+                    'timestamp': timestamp,
+                    'name': _get_covid_test(test).OBSERVATION_NAME,
+                    'value': result_string
+                }
+            )
+
+        ticker += data
+
+    ticker = sorted(ticker, key=lambda i: i['timestamp'])
```
fredkingham: This looks wrong, if we have 30/05/2020, 31/05/2020, 01/06/2020 it will reorder them 01/06/2020, 30/05/2020, 31/05/2020 no?

plugins/covid/lab.py
```diff
@@ -115,16 +175,24 @@ def get_result(test):
...
     Given the plugins.labtests.models.LabTest instance TEST, which is a
-    COVID 19 test, return the date of the relevant observation
+    COVID 19 test, return the datetime of the relevant observation
     """
     covid_test = _get_covid_test(test)
     observations = test.observation_set.filter(
         observation_name=covid_test.OBSERVATION_NAME
     )
-    return observations[0].observation_datetime.date()
+    return observations[0].observation_datetime
```
fredkingham, line 187: Do all observations definitely have a datetime? Logically they should but we can't guarantee it.

davidmiller: 0.07% of tests have a null datetime. 
100% of these also have a null observation_name

plugins/covid/lab.py
davidmiller: it has to get to 3 to pass the > 2 test on line 233

plugins/covid/loader.py
davidmiller: In practice this doesn't seem to be happening in the logs.
Even if it did, we effectively retry an hour later anyway so meh.

PR #971, opened 2020-06-03
Vanity metrics

davidmiller:


elcid/views.py
```diff
@@ -260,3 +260,20 @@ def post(self, *args, **kwargs):
                 reason=models.ChronicAntifungal.DISPENSARY_REPORT
             )
         return HttpResponseRedirect(self.get_redirect_url())
+
+
+class ElcidDashboardView(LoginRequiredMixin, TemplateView):
+    template_name = 'elcid/dashboard.html'
+
+    def get_context_data(self, *a, **k):
+        from plugins.labtests import models as labmodels
```
fredkingham, line 269: Does this need to be here? (I'm guessing it does, but can't figure out why)

PR #977, opened 2020-06-08
Demographics fields

fredkingham:


fredkingham:
Things to note.

1. **Other subrecords** ContactDetails, GP and NextOfKin are already subrecords. (CD and NOK are in TB). Also TB has language in Communication. models. I made the call to just bring the information in as seperate subrecords. We can look at merging later but mixing un editable and editable fields felt like something we would not want to deal with for the time being.

2. **Potential issues** Notable we are using the last updated/date inserted timestamps. Also notable the upstream view contains a foreign key table to a GP table. It is not clear that when the GP table get's updated the last updated/inserted time stamp will change. I've decided to assume that it will for the time being because there is an FK ID and historically their method of updating is to dump and rebuild tables so I'm assuming that the FK ID will change and therefore the time stamp will change but I cannot verify this at this time.

3. **Flow** We are stopping the lab batch touching demographics, the `sync_patient` as called by the `sync_demographics` management command will update all of the patients. The patients information will be loaded in when they are added by `load_patient`.  An initial run of the sync patients  management command should set us up to be good to go.

4. **Other** I have made them subrecords, mostly because they are singletons and this _seems fair_ as unlike lab tests/appointments you are never going to have 1000s of them. They are currently excluded from the extract because the information is patient identifying. They do not have form templates as they should not be updated.

PR #981, opened 2020-06-10
Dischargesummaries

davidmiller:


plugins/dischargesummary/loader.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    api = ProdAPI()
+
+    demographic = patient.demographics()
+
+    summary_count = patient.dischargesummaries.count()
+
+    summaries = api.execute_hospital_query(
+        Q_GET_SUMMARIES,
+        params={'mrn': demographic.hospital_number}
```
fredkingham, line 56: We have had cases where hospital number is None/empty string (even though it never should be), probably should check that.

PR #1005, opened 2020-08-18
Covid calculator flush

fredkingham:
We had an issue where a user looked at the calculator dashboard while it was being calculated.

Previously this would throw an error because the dashboard did not exist.

The job ran once an hour and too just over 4 mins.

This changes it so all db deletes/writes are done in a single function and that function is atomic.

The deletes/writes take 0.35 seconds, and the transaction means that the database will not be read in an inconsistent state.


davidmiller:
Is there a reason we can't put the transaction on the existing `calculate` method?

fredkingham:
So we can. I thought I'd make it more efficient while making the change.

PR #1006, opened 2020-08-18
Moves the print style sheets for letters into the elcid css as plugin…

fredkingham:
…s css in opal 0.17.1 are screen only


fredkingham:
closing this, it doesn't matter, the application style sheet is also print only so this is pointless.

PR #1020, opened 2020-09-21
Run cron job writers at the end of the deploy run. rm /etc/cron.d/elc…

davidmiller:
…id* at the start

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -869,6 +869,8 @@ def _deploy(new_branch, backup_name=None, remove_existing=False):
     if backup_name and not os.path.isfile(backup_name):
         raise ValueError("unable to find backup {}".format(backup_name))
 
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -869,6 +869,8 @@ def _deploy(new_branch, backup_name=None, remove_existing=False):
     if backup_name and not os.path.isfile(backup_name):
         raise ValueError("unable to find backup {}".format(backup_name))
 
+    local("sudo rm /etc/cron.d/elcid*")
```
fredkingham, line 872: Works for everything apart from write_cron_lab_tests

I guess that's enough but shouldn't we change write_cron_lab_tests to put in the project name as well.

Also shouldn't this do ```local("sudo rm /etc/cron.d/{}*".format(PROJECT_NAME))```

PR #1033, opened 2020-10-16
Covid follow up episode creator

davidmiller:


plugins/covid/management/commands/create_covid_episodes.py
```diff
@@ -3,11 +3,20 @@
...
 
 
 class Command(BaseCommand):
 
     def handle(self, *args, **options):
         loader.create_followup_episodes()
+
+        without_followups = Patient.objects.filter(
+            covid_patient__date_first_positive__isnull=False
+        ).exclude(episode__category_name=CovidEpisode.display_name).count()
```
fredkingham: I think this will return an int from .count()

PR #1036, opened 2020-10-20
Disk space errors

davidmiller:


fredkingham:
Do you want to wrap the whole thing in a try catch with a logger .error that emails us if something is wrong?

Arguably its over defensive but it gives us confidence that it actually works.

PR #1047, opened 2020-11-23
Remove proxy settings from deployment

davidmiller:


fredkingham:
Are we removing proxy settings forever?

In which case do we want to remove
https://github.com/openhealthcare/elcid-rfh/blob/8bc35ba8d205e781cd8de15922914b0cc698ebf9/fabfile.py#L6
https://github.com/openhealthcare/elcid-rfh/blob/8bc35ba8d205e781cd8de15922914b0cc698ebf9/fabfile.py#L889

PR #1057, opened 2020-12-21
Remove discharge summaries and make ipls fail

fredkingham:
Sorry this looks bigger than it is because we move @transaction.atomic into a with block. This means that if an InitialPatientLoad fails it gets stored in the db as a failure.

It also added logger.error so that we get an email when it fails.

It removes the test case, this needs more work and potentially some debate about how we handle apis.

It also removes the discharge summary load for the time being.


fredkingham:
Closing this, we're going in a different direction.

PR #1058, opened 2020-12-22
Use the get_api in the loaders

fredkingham:
So this is how I initially envisioned the api would work, the idea being you could switch out the api to get realistic data for when you were in a dev enviroment.

I'm fully open to other options.

This fixes the unit tests and makes them work like the intrahospital_api.loader


fredkingham:
I think we are going to look for ways that don't use `get_api`

PR #1060, opened 2020-12-22
bump the travis version to 3.6 as we don't use 3.5 in prod (or indeed…

fredkingham:
… anywhere)


fredkingham:
I'll put this in a different PR

PR #1111, opened 2021-04-02
Rfh barnett split

fredkingham:
includes https://github.com/openhealthcare/elcid-rfh/pull/1108

so this may be contraversial... We store all numbers for each location.

The question was how do you deal with a patient who is covid positive at RFH and then subsequently covid positive at Barnet. Is that a new patient, which date do we use?

My answer is that for the individual views its a snap shot for that hospital, as a patient may be covid positive at the St Thomas' and then covid positive at Barnet, except in that case we would not know so we don't special case it, we look at them all seperately. 

It means the load takes twice as long but that's still 20% of what is used to so...

Screen Shot 2021-04-02 at 18 31 32


fredkingham:
closing in favour of https://github.com/openhealthcare/elcid-rfh/pull/1113

PR #1120, opened 2021-05-06
IPC module - initial Alpha

davidmiller:


plugins/ipc/lab.py
```diff
@@ -0,0 +1,174 @@
...
+    def __repr__(self):
+        return "\n".join(
+            [f"{self.test.lab_number} {self.test.test_code} {self.test.test_name} {self.test.datetime_ordered}"] +
+            [f"{o.observation_name} {o.observation_value}" for o in self.get_observations()]
+        )
+
+    def get_observations(self):
+        return self.test.observation_set.filter(
+            observation_name__in=self.OBSERVATION_NAMES
+        ).exclude(observation_value='.{}')
```
fredkingham, line 30: I think you want to also exclude if observation = None as most of the `def alert` assume the existence of a string.

We have observations where the value is None, I'm guessing it should not happen but it _could_

davidmiller: Observations of these codes and names  ? 

Surely if this was the case running `manage.py ipc` would stack trace?

plugins/ipc/lab.py
```diff
@@ -0,0 +1,174 @@
...
+    TEST_CODE         = 'DIFF'
+    OBSERVATION_NAMES = ['C.diff PCR test', 'C.diff Toxin test']
+    ALERT_CATEGORY    = models.InfectionAlert.CDIFF
+
+    def alert(self):
+        """
+        If either observation is Positive, return True
+        """
+        for observation in self.get_observations():
+            if observation.observation_value == 'POSITIVE':
```
fredkingham, line 68: Are you sure we don't want observation.observation_value.startswith('POSITIVE')?

plugins/ipc/utils.py
```diff
@@ -0,0 +1,22 @@
...
+def alphanum_key(s):
+    """ Turn a string into a list of string and number chunks.
+        "z23a" -> ["z", 23, "a"]
+    """
+    return [ tryint(c) for c in re.split('([0-9]+)', s) ]
+
+def sort_nicely(l):
+    """ Sort the given list in the way that humans expect.
+    """
+    return sorted(l, key=alphanum_key)
```
fredkingham, line 26: This feels like it might break with an int str error, ie

a = ["z", 23, "a"]
b = [23, "z", "a"]
sorted([a, b]) -> raises an error

plugins/ipc/views.py
```diff
@@ -0,0 +1,126 @@
...
+        mrsas  = ['MRSA'] + [monthly[k][models.InfectionAlert.MRSA] for k in sorted(monthly.keys())[-12:]]
+        tbs    = ['TB'] + [monthly[k][models.InfectionAlert.TB] for k in sorted(monthly.keys())[-12:]]
+        vres   = ['VRE'] + [monthly[k][models.InfectionAlert.VRE] for k in sorted(monthly.keys())[-12:]]
+
+        context['overview_data'] = [ticks, cdiffs, cpes, mrsas, tbs, vres]
+
+        context['rfh_patients'] = UpstreamLocation.objects.filter(building='RFH').count()
+        context['rfh_siderooms'] = UpstreamLocation.objects.filter(building='RFH', room__startswith='SR').count()
+        context['weekly_alerts'] = models.InfectionAlert.objects.filter(
+            trigger_datetime__gte=today-datetime.timedelta(days=7)).count()
```
fredkingham, line 45: I may be reading this wrong, do we not just care about RFH Infection alerts?

plugins/admissions/management/commands/parse_admissions.py
```diff
@@ -0,0 +1,44 @@
...
+                room=encounter.pv1_3_room,
+                bed=encounter.pv1_3_bed
+            )
+            location.hospital  = encounter.pv1_3_hospital
+            location.patient   = encounter.patient
+            location.admitted  = encounter.pv1_44_admit_date_time
+            location.encounter = encounter
+
+            location.save()
+            self.stdout.write('.')
```
fredkingham, line 44: you probably don't need this

plugins/ipc/lab.py
davidmiller: Possibly, but we're reviewing this with clinical team later this week - right now these are indicative only

plugins/ipc/templates/records/infection_alert.html
```diff
@@ -0,0 +1,13 @@
...
+Lab #: [[ item.lab_test_id ]]
+<span ng-show="item.seen">
+     Seen
+   <br />
+</span>
+<span ng-hide="item.seen">
+  Unseen
+</span>
+<span ng-show="item.comments">
+   <br />
```
fredkingham, line 11: If a patient is seen they have a `<br>` after but if there are comments they have a `<br>` before, I think you just want `Unseen <br />`

plugins/ipc/utils.py
davidmiller: I was literally just copy and pasting Ned Batchelor https://nedbatchelder.com/blog/200712/human_sorting.html

Also, why do you think that raises an error? 

```
>>> a = ["z", 23, "a"]
>>> b = [23, "z", "a"]
>>> sorted([a, b])
[[23, 'z', 'a'], ['z', 23, 'a']]
>>>
```

plugins/ipc/views.py
davidmiller: > do we not just care about RFH Infection alerts?

Can you rephrase the question?
I don't know what you're asking :)

plugins/ipc/templatetags/ipc.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+from django import template
+
+from plugins.admissions.models import UpstreamLocation
+
+from plugins.ipc.episode_categories import IPCEpisode
+
+register = template.Library()
+
+@register.inclusion_tag('templatetags/ward_count.html')
```
fredkingham, line 12: This is probably a filter rather than a tag (doesn't really matter)

PR #1121, opened 2021-05-11
Minor change so that the left menu does not scroll instead is always …

fredkingham:
…shown in its entirety

rather than scroll the left menu let it scroll with the page and always show it.


fredkingham:
closing in favour of https://github.com/openhealthcare/elcid-rfh/pull/1122

PR #1122, opened 2021-05-11
Change the scroll bar to be custom styled

fredkingham:
Screen Shot 2021-05-11 at 14 48 42


fredkingham:
So it turns out the scroll bar only shows if you have an external mouse... This changes the styling of the scroll bar, not sure if you think its worth it (it is for me...)

The scroll bar only appears on hover or touch devices like slacks.

PR #1127, opened 2021-05-12
Patient API speed

davidmiller:
20% bump on an indicative patient by not bothering with subrecords for categories this patient is not relevant to.


davidmiller:
We're not going to do this any time soon...

PR #1129, opened 2021-05-13
Bulk create speed up 1

fredkingham:
* Changes the creation of lab tests by the batch load to use bulk_create on the observations.
* Skip lab tests when we are not provided a hospital number.
* Skip lab tests when we are not provided either a lab number or a test name.

### DIFF
--- a/intrahospital_api/management/commands/batch_load2.py
+++ b/intrahospital_api/management/commands/batch_load2.py
@@ -21,7 +21,6 @@ from plugins.monitoring.models import Fact
 def send_email(title, template_name, context):
     html_message = render_to_string(template_name, context)
     plain_message = strip_tags(html_message)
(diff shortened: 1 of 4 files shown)


intrahospital_api/management/commands/batch_load2.py
```diff
@@ -63,13 +66,15 @@ def handle(self, *args, **options):
...
+                continue
+
             if not patient_demographics_set.exists():
                 continue # Not in our cohort
 
-            update_lab_tests.update_tests(
-                patient_demographics_set.first().patient,
-                item["lab_tests"],
-            )
+            item['lab_tests'] = [i for i in item['lab_tests'] if i["test_name"] or i["external_identifier"]]
```
davidmiller: Can we make this _much_ clearer e.g. definitely not in a list comprehension, and with comments

PR #1130, opened 2021-05-14
Give the remove button more room so that it does not go over the edge…

fredkingham:
… of the card


fredkingham:
was
Screen Shot 2021-05-13 at 20 55 21




now
Screen Shot 2021-05-14 at 13 27 03

PR #1137, opened 2021-05-19
Tb list improvements

fredkingham:
visual gap between doctor and nurses appointments
dates have a white background with a brown border
time renders as a 24 hour clock
removes cancelled appointments
Screen Shot 2021-05-19 at 14 39 44


plugins/tb/templates/tb/clinic_list.html
```diff
@@ -31,9 +31,24 @@ <h3 class="content-offset left-offset-10">
...
+            <td></td>
+            <td></td>
+          </tr>
+        {% endif %}
+      {% endifchanged %}
+    {% endif %}
     <tr>
       <td>
-        {{ appointment.start_datetime.time }}
+        {{ appointment.start_datetime.time|date:"G:i" }}
```
davidmiller: do we need to go to `.time` ?

PR #1154, opened 2021-06-04
We no longer use the external identifier to track whether something h…

fredkingham:
…as been updated externally, instead we use the master file meta


fredkingham:
we are going to use this, just update the updater

PR #1155, opened 2021-06-04
Make sure we set the unix user in the sync demographics cron job

fredkingham:


fredkingham:
This is already included in the demographics sync time PR

PR #1165, opened 2021-06-24
Set the external system if we update a patient subrecord in the updat…

fredkingham:
…e demographics job


fredkingham:
External system is used by the reconcile demographics and by the demographics modal template to decide whether the model is updatable.

Lets set this.

PR #1166, opened 2021-06-24
Imaging synch

fredkingham:


fredkingham:
This job now takes less than 2 mins to run, so I've changed it to run twice an hour.

PR #1172, opened 2021-07-07
Adds temperature into the observation panel and form

fredkingham:


fredkingham:
This automatically brings it into the nurses letter

PR #1173, opened 2021-07-08
Add a heavily tested file that parses lab tests which contain organisms

fredkingham:
This parses the lab test organisms result. I think it could go into a get_organisms function on lab test but wanted to get feed back. 

I'm using it on the prototype MDT list but thought it was seperate enough to get evaluated as its own thing.


fredkingham:
This does not include `U` and we need to find out what these refer to

fredkingham:
It also does not include `W`

PR #1207, opened 2021-10-01
In the tb clinic list if there are only cancelled appointments on a d…

fredkingham:
…ate, make sure we don't error


fredkingham:
closing in favour of #1208 which goes to the correct branch

PR #1213, opened 2021-10-04
Admissions duplicate patients

fredkingham:
* If there are encounters for duplicate patients we should load the encounters in on both the patients, in keeping with the other loaders.

* Changes the load recent encounters method to be load encounters since. This is so if the upstream server connection fails for whatever reason we can manage the load in of the data.

* Make sure we update the has_encounter status for the patients we load in.


fredkingham:
closing as we're going to check the behaviour with duplicate patients, included the other commits on https://github.com/openhealthcare/elcid-rfh/pull/1216

PR #1226, opened 2021-10-13
Update opal

fredkingham:


fredkingham:
so this _should_ work.

If it is successful on test after a week, I suggest we pypi release opal

Other things to check
- [ ] test the email error sending
- [ ] test resend password
- [ ] extract

PR #1227, opened 2021-10-14
Lab detail

fredkingham:
This is kinda for discussion. I think its quite good. The main problem with it is load time which is long because there is no index in lab number and we're not looking at a specific patient


davidmiller:
Merged into v102

PR #1241, opened 2021-11-01
Make the empty patient list no go over the brown seperating line betw…

fredkingham:
…een the title and the page

turns
Screen Shot 2021-11-01 at 14 51 32
into

Screen Shot 2021-11-01 at 14 52 01


davidmiller:
Merged into v102

PR #1261, opened 2021-12-20
Ipc advice templates btns

fredkingham:
## This is pointing to the wrong branch, just not sure what the right branch is
Screenshot 2021-12-20 at 13 01 45
Screenshot 2021-12-20 at 13 01 52
Screenshot 2021-12-20 at 13 02 04


fredkingham:
We don't use ___ as these are markdown syntax so it get's weird if there are multiple uses

davidmiller:
Everything useful in here already in the release branch as far as I can tell, with subsequent fixes.

PR #1279, opened 2022-01-19
Adds some additional fields to the lung function test and shows them …

fredkingham:
…in the tb detail page in a similar way as we do the observations


fredkingham:
Screenshot 2022-01-19 at 18 09 45

PR #1310, opened 2022-10-11
Reject everything that is not heading to our expected address

fredkingham:
So we want to make sure that if people are pen testing our server we reject their requests and don't get emails about it.

The below works for the following test cases on test
`curl -i elcid-rfh-test` get's me the 302 you'd expect

`curl -i http://elcid-rfh-test-1 --resolve 'elcid-rfh-test-1:80:192.168.21.203'` now returns a 403


fredkingham:
closing this in favour for https://github.com/openhealthcare/elcid-rfh/pull/1315

PR #1316, opened 2022-10-24
Copies the new deny-others logic onto the ansible deployment branch

fredkingham:
On prod and private we are putting in a change so that if the users do not match the server name we expect we 403.

This does that logic onto the ansible deployment branch.

It also 
* fixes the log format in nginx so that its the same as we currently use in prod
* corrects the servernames to what we expect on live


fredkingham:
This is pending testing by the users

PR #1324, opened 2022-11-16
Adds a default timeout to our execute queries

fredkingham:
The default pytds timeout is infinite. Locally running queries sometimes they take a significant amount of time to make the connection. This makes sure that we timeout and don't run forever.


fredkingham:
Removing this and having an issue

fredkingham:
Closing this and raising it as an issue.

PR #1325, opened 2022-11-18
Zero fixes for loaders where the upstream system does not use zero prefixes

fredkingham:
This fixes all loaders where the upstream does not use prefixed zeros. In this case fixes means that the loaders will save for all patients whether prefixed by zeros or if they are duplicates.

This means
* When loading for a specific patient we strip off the leading zeros.
* When loading since a certain time we update all patients if they are prefixed with zeros or not.

It uses the `hospital_numbers_to_patients` which returns a list of patients for a hospital number. This is when we have multiple patients with the same hospital number or patients where their hospital number without preceding zeros is the same.

The plan would be to change `hospital_numbers_to_patients` to account for merged patients in the future.

This can be released now. We should rerun the loaders for all patients with prefixed zeros.

### DIFF
--- a/intrahospital_api/loader.py
+++ b/intrahospital_api/loader.py
@@ -4,6 +4,7 @@ Functions for loading data from upstream.
 import datetime
 import traceback
 import json
+from collections import defaultdict
 
 from django.db import transaction
 from django.utils import timezone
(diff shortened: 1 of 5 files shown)


fredkingham:
We are closing this as we're going act as if 0 prefixes should never exist within elcid.

PR #1326, opened 2022-11-18
No more add patient with zero prefixes

fredkingham:
When we search for a patient in the add patient flow, match any patient if they have prefixed zeros. e.g. if you try and add 193 and 00193 exists in elcid then match 00193. This is to stop duplicates being created in our system.

If no patient is matched and we are going to create a patient then make sure we strip off the prefixed zeros before creating them. We should not be creating any new patients with prefixed zeros.


fredkingham:
We are closing this as we're going act as if 0 prefixes should never exist within elcid.

PR #1332, opened 2022-11-21
Lab test leading zeros

fredkingham:
1. Make sure that when saving lab tests from we strip off prefixed zeros from hospital numbers
2. Make sure when querying by hospital number for lab tests upstream we check to see if there are any lab tests attached to the hn but where the hn has preceding zeros.

Note we use an in query because it takes ~13 seconds, if we do it in serial and aggregate  takes ~20 seconds


fredkingham:
Changing this in favour of https://github.com/openhealthcare/elcid-rfh/pull/1338

PR #1334, opened 2022-11-23
Bring back the old labs model to delete the underlying db tables

fredkingham:
On production we have stale tables left over from when we deleted the labs module. This means that you get a postgres error when trying to delete a patient who has some of the old lab tests as it violates a foreign key constraint.

This brings back the module so that we have a migration that will delete the old database tables. After the migrations are run on production we can remove this app again.


fredkingham:
closing this, we're going to use psql instead

PR #1339, opened 2022-11-25
135 write to epr initials

fredkingham:
EPR requires a written_by field.

In TB patient consulations and other services microbiology input this is populated by the initials field. 

We change the timeline to accept standard opal template tag validation, in this case required=True. This means changing it so that after the form is submitted the form is marked as pristine. In angular terms this is not submitted and not dirty. This is so validation is reset for the next time the form is submitted.

With the covid branch its simpler. We populated the EPR initials field with the consulation from CovidFollowUpCall, CovidFollowUpCallFollowUpCall, CovidSixMonthFollowUp. We make this field required in the form, not in the model as we have existing data with None in the clinican.

Note the logic in the RNOH form is unchanged but because the API for the clinicalTimeline has change (save takes a form now), the template has been changed.

### DIFF
--- a/elcid/templates/inline_forms/clinical_timeline.html
+++ b/elcid/templates/inline_forms/clinical_timeline.html
@@ -72,6 +72,6 @@
 {% endblock %}
 
 {% block send_upstream_button %}
-<i ng-show="item.reason_for_interaction && item.reason_for_interaction.indexOf('RNOH') !== 0" class="fa fa-exchange edit pull-right pointer"
+<i ng-show="item.reason_for_interaction && item.reason_for_interaction.indexOf('RNOH') === -1" class="fa fa-exchange edit pull-right pointer"
    ng-click="open_modal('SendUpstreamCtrl', '/templates/send_upstream_modal.html', {patient:patient, item:item, refresh_patient:refresh, refresh_timeline:clinicalTimeline.getClinicalAdvice})"></i>
 {% endblock %}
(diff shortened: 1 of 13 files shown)


elcid/templates/inline_forms/clinical_timeline.html
```diff
@@ -72,6 +72,6 @@
 {% endblock %}
 
 {% block send_upstream_button %}
-<i class="fa fa-exchange edit pull-right pointer"
+<i ng-show="item.reason_for_interaction && item.reason_for_interaction.indexOf('RNOH') !== 0" class="fa fa-exchange edit pull-right pointer"
```
davidmiller: Wouldn't a `-1` spelling of this be safer against future spellings of reason for interaction?

fredkingham: I thought I was future proofing by making sure we didn't catch false positives. 

I was wary we might have a RNOH in the middle of a non RNOH reason for interaction (e.g. discharge not for patient going to RNOH). I'm happy either way.

davidmiller:
Think we need to show the error by the save button - currently easy for it to be invalid but show no indication on screen...

Screenshot 2023-01-11 at 13 30 15

fredkingham:
good call, this is done and deployed to the test server.
Screenshot 2023-01-12 at 11 46 35

Screenshot 2023-01-12 at 11 46 54

PR #1342, opened 2022-11-29
Bug fix the create_rfh_patient loader method to read the async parameter

fredkingham:
At the moment if you pass run_async to create_rfh_patient_from_hospital_number it will never be used.

This fixes that behaviour.

### DIFF
--- a/intrahospital_api/loader.py
+++ b/intrahospital_api/loader.py
@@ -103,11 +103,7 @@ def create_rfh_patient_from_hospital_number(hospital_number, episode_category, r
         category_name=episode_category.display_name,
         start=datetime.date.today()
     )
-
-    if run_async is None:
-        load_patient(patient, run_async=run_async)
-    else:
-        load_patient(patient)
+    load_patient(patient, run_async=run_async)
     return patient
 
 


intrahospital_api/loader.py
```diff
@@ -104,7 +104,7 @@ def create_rfh_patient_from_hospital_number(hospital_number, episode_category, r
         start=datetime.date.today()
     )
 
-    if run_async is None:
+    if run_async is not None:
```
davidmiller: Doesn't this run it synchronously if the spelling is `create_rfh_patient_from_hospital_number('213', 'infection', run_async=False)`
That also feels like not the right thing to do

fredkingham: So as written, it is not possible to use the `run_async` parameter.

I want to be able to say `run_async=False` if I want it to run synchronously.

If you think you should never be able to run it synchronously then we should remove the paramenter, but I think its a good feature.

davidmiller: Why doesn't it just pass through the parameter whatever the value, why the is not None check?

PR #1344, opened 2022-11-30
Update original mrn

fredkingham:
Depends on https://github.com/openhealthcare/elcid-rfh/pull/1343.

We vary from the original requirements by using the update_from_dict method rather than the save method.

This is because update_from_dict should only be called from the browser, therefore we can differentiate between models updated by the browser vs models updated by the backend.

When updated by the browser the original_mrn should always be set to None as old MRN will no longer be visible in elcid.


fredkingham:
closing in favour of https://github.com/openhealthcare/elcid-rfh/pull/1346

fredkingham:
closing in favour of https://github.com/openhealthcare/elcid-rfh/pull/1348

PR #1345, opened 2022-11-30
Merge search updates

fredkingham:
Depends on https://github.com/openhealthcare/elcid-rfh/pull/1343.

Changes it so that we include merged MRNs in a patients query results (ie if the patient MRN 123 was previously associated with MRN 234, a search result for 234 returns patient 123)

Changes it so that when we display the result we display the previously associated MRNs.

Screenshot 2022-11-30 at 21 41 13


fredkingham:
closing in favour of https://github.com/openhealthcare/elcid-rfh/pull/1347

PR #1352, opened 2022-12-02
Merge nationality pathway

fredkingham:
This PR makes sure the nationality pathway displays the previous mrn in the footer and also removes the previous mrn on save.

Things to note...
1. The update_from_dict update...
So without this the tests fail, because in the case that you remove the previous_mrn and then call update_from_dict, the update_from_dict method saves the previous_mrn. In many ways this is an edge case however the situation is that the previous_mrn should _never_ be populated from the front end and we should always zero out the previous mrn on save. So this PR includes that.

2. Demographics
So this pathway updates a single field on Demographics. We ignore demographics.previous_mrn (on display and update) we just use the birth_place field which is not touched by the upstream api.

so lets review what this means,

A patient that previously has demographics (manually editted)
It will not show that a birth place previously added under another MRN as editted under another MRN. Now arguably this does not matter birth place is never added by upstream MRNs so it should not really matter which MRN it was added under. Alternatively I would suggest we move birth_place to nationality. Migrate the data to nationality, remove it from the demographics form and remove it from advanced search. Your call.

### DIFF
--- a/plugins/tb/pathways.py
+++ b/plugins/tb/pathways.py
@@ -211,6 +211,19 @@ class NationalityAndLanguage(pathways.PagePathway):
             client_demographics = data.pop("demographics")
             our_demographics.birth_place = client_demographics[0]["birth_place"]
             our_demographics.save()
-        return super().save(
+
+            # We don't necessarily call update_from_dict on
+            # pathway subrecords, only if they are editted,
+            # so force removal of previous_mrn if we hit
+            # the save method.
+            patient.communinicationconsiderations_set.update(
+                previous_mrn=None
+            )
+            patient.nationality_set.update(
+                previous_mrn=None
+            )
+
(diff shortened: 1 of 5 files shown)


plugins/tb/pathways.py
```diff
@@ -211,6 +211,26 @@ def save(self, data, user=None, episode=None, patient=None):
...
+            # so force removal of previous_mrn if we hit
+            # the save method.
+            patient.communinicationconsiderations_set.update(
+                previous_mrn=None
+            )
+            patient.nationality_set.update(
+                previous_mrn=None
+            )
+        # print('before')
+        # print(f'coms {patient.communinicationconsiderations_set.get().previous_mrn}')
```
davidmiller: Guess we don't still need these?

PR #1356, opened 2022-12-05
Makes bed status check for merged MRNs

fredkingham:
When saving bed statuses, save them to the correct patient/MRN if that MRN has been merged.

### DIFF
--- a/plugins/admissions/loader.py
+++ b/plugins/admissions/loader.py
@@ -357,11 +357,16 @@ def load_bed_status():
                     demographics__hospital_number=bed_status.local_patient_identifier
                 ).first()
 
+                if not patient:
+                    patient = Patient.objects.filter(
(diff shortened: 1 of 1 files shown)


davidmiller:
I can imagine wanting this to read 

```python 
from elcid.utils import get_or_create_patient
#...
mrn = bed_status.local_patient_identifier
bed_status.patient = get_or_create_patient(mrn, episode_category=InfectionService, run_async=True)
```

If only because we're now going to want to run exactly this check for Demographics.hospital_number, MergedMRN.mrn, create_patient all over the place and it gives us a consistent API

fredkingham:
We can but we don't do that with zero prefixes, we raise an error. I'm happy either way but given we went one way with zero prefixes...

fredkingham:
https://github.com/openhealthcare/elcid-rfh/pull/1361

PR #1357, opened 2022-12-06
Changes the lab test load to also load lab tests for any merged MRNs

fredkingham:
If an inactive MRN has been merged into an MRN. Make sure we query lab tests with the inactive MRN when looking for lab tests for the active MRN.

### DIFF
--- a/intrahospital_api/apis/prod_api.py
+++ b/intrahospital_api/apis/prod_api.py
@@ -808,18 +808,28 @@ class ProdApi(base_api.BaseApi):
     @timing
     def results_for_hospital_number(self, hospital_number):
         """
-        returns all the results for an MRN
+        Returns all the results for an MRN
         aggregated into labtest: observations([])
+
+        Also checks for alternative MRNs to fetch, due to
+        upstream 0 prefixing or Cerner merges.
         """
-        raw_rows = []
         merged_mrns = list(set(MergedMRN.objects.filter(
             patient__demographics__hospital_number=hospital_number
         ).values_list('mrn', flat=True)))
+
         all_mrns = [hospital_number] + merged_mrns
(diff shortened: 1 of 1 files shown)


intrahospital_api/apis/prod_api.py
```diff
@@ -811,9 +811,14 @@ def results_for_hospital_number(self, hospital_number):
...
         raw_rows = self.raw_data(hospital_number)
-        for other_hn in other_hns:
-            raw_rows.extend(self.raw_data(other_hn))
+        for zero_prefixed_mrn in zero_prefixed_mrns:
+            raw_rows.extend(self.raw_data(zero_prefixed_mrn))
+        merged_mrns = list(set(MergedMRN.objects.filter(
+            patient__demographics__hospital_number=hospital_number
+        ).values_list('mrn', flat=True)))
+        for merged_mrn in merged_mrns:
+            raw_rows.extend(self.raw_data(merged_mrn))
```
davidmiller: Why can't merged MRNs have 0-prefixed lab test identifier entries?

fredkingham: The cerner master file implicitly.

So the logic we're using is. A patient is merged if they exist in the cerner master file as inactive and have a merge to another patient. As the cerner master file contains no zero prefixes, a merged MRN can't have a zero prefix.

fredkingham: actually after conversation. Good call, done.

davidmiller:
I think I'd do it this way: https://github.com/openhealthcare/elcid-rfh/pull/1362

fredkingham:
> I think I'd do it this way: #1362

Sure if you think that's easier to read.

PR #1360, opened 2022-12-06
Merge batch lab tests

fredkingham:
If the upstream lab test table is using inactive MRNs, add the lab tests to the correct active MRN for that table.

While doing so remove some of the old IPL logic as this is inconsistently applied compared with other loaders.

### DIFF
--- a/intrahospital_api/loader.py
+++ b/intrahospital_api/loader.py
@@ -325,18 +325,9 @@ def update_patient_from_batch(demographics_set, data_delta):
 
 @timing
 def update_from_batch(data_deltas):
-    # only look at patients that have been reconciled
-    demographics_set = emodels.Demographics.objects.filter(
-        external_system=EXTERNAL_SYSTEM
-    )
-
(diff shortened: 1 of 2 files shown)


fredkingham:
I'm going to move this back into in progress as I think we should use `get_patient` from https://github.com/openhealthcare/elcid-rfh/pull/1361

intrahospital_api/loader.py
```diff
@@ -296,44 +296,38 @@ def _batch_load():
 @transaction.atomic
 def update_patient_from_batch(demographics_set, data_delta):
```
davidmiller, line 297: Are you sure this is the function that actually runs in production?

fredkingham: good call, closing this for https://github.com/openhealthcare/elcid-rfh/pull/1390

fredkingham:
closing for https://github.com/openhealthcare/elcid-rfh/pull/1390

PR #1361, opened 2022-12-06
Adds the get_or_create_patient function that looks for patients by MRN

fredkingham:
It looks by hosptial number locally, then by merged MRN locally.

If its not found it checks upstream whether this hospital number has been marked as in active and merged. If it has it returns the active MRN that it is merged with.


intrahospital_api/loader.py
```diff
@@ -99,16 +101,54 @@ def create_rfh_patient_from_hospital_number(hospital_number, episode_category, r
...
+        return patient
+
+    active_row = actives[0]
+    patient = Patient.objects.create()
+    patient.demographics_set.update(
+        hospital_number=active_row["PATIENT_NUMBER"]
     )
+    merged_mrns = []
+    for row in rows:
+        if row["ACTIVE_INACTIVE"] == emodels.MergedMRN.INACTIVE:
```
davidmiller: Does this work ? I thought this would give you the active MRN not the merged one ?

intrahospital_api/loader.py
```diff
@@ -99,16 +101,54 @@ def create_rfh_patient_from_hospital_number(hospital_number, episode_category, r
...
+            _, merge_dt = update_demographics.get_mrn_and_date_from_merge_comment(
+                row["MERGE_COMMENTS"]
+            )[0]
+            merged_mrns.append(emodels.MergedMRN(
+                patient = patient,
+                mrn=row["PATIENT_NUMBER"],
+                upstream_merge_datetime=merge_dt,
+                merge_comments=row["MERGE_COMMENTS"]
+            ))
+    emodels.MergedMRN.objects.bulk_create(merged_mrns)
```
davidmiller: "There are never more than 3" <- this would be an example of a time I think it's not worth the container datastructure to bulk create

elcid/utils.py
```diff
@@ -81,11 +82,30 @@ def natural_keys(text):
     return [ atoi(c) for c in re.split(r'(\d+)', text) ]
 
 
+def get_or_create_patient(mrn, episode_category, run_async=None):
```
davidmiller: Where do we think this is going to be called?

elcid/utils.py
```diff
@@ -81,11 +82,30 @@ def natural_keys(text):
...
+    from intrahospital_api import loader
```
davidmiller: Honestly I'd be tempted to put this in the intrahospital_api if we're doing this kind of thing

intrahospital_api/update_demographics.py
```diff
@@ -194,6 +195,69 @@ def update_if_changed(instance, update_dict):
...
+    for match in found:
+        mrn = match[0]
+        date_str = f"{match[2]} {match[1]} {match[3]} {match[4]}{match[5]}"
+        merge_dt = datetime.datetime.strptime(date_str, "%d %b %Y %I:%M%p")
+        result.append((mrn, timezone.make_aware(merge_dt),))
+    # return by merged date
+    return sorted(result, key=lambda x: x[1], reverse=True)
+
+
+def get_related_rows_for_mrn(mrn):
```
davidmiller: Is this really what you want? Is there a reason I need more than `fn(mrn)` ... `return primary, [merged_mrn, merged_mrn,...]` where primary and merged_mrn are character strings with the MRN value?

intrahospital_api/update_demographics.py
```diff
@@ -194,6 +195,103 @@ def update_if_changed(instance, update_dict):
...
+
+    The merged comments can be nested for for MRN x
+    we can have MERGE_COMMENT "Merged with y on 21 Jan"
+    Then for y we can have the merge comment "Merged with z on 30 Mar"
+    This will return the rows for x, y and z
+
+    If there are no related rows, ie the patient
+    is not merged, return None.
+    """
+    query = """
```
davidmiller: Could be convinced otherwise, but I _think_ I quite like the convention of queries all together at the top of the file in loaders as global variables. I think it's running debugging queries where it's particularly helpful

intrahospital_api/update_demographics.py
```diff
@@ -194,6 +195,103 @@ def update_if_changed(instance, update_dict):
...
+    related_mrns = [i[0] for i in get_mrn_and_date_from_merge_comment(query_result["MERGE_COMMENTS"])]
+    for related_mrn in related_mrns:
+        related_result = api.execute_hospital_query(
+            query, {"mrn": related_mrn}
+        )
+        mrn_to_row[related_mrn] = related_result
+        related_related_mrns = [
+            i[0] for i in get_mrn_and_date_from_merge_comment(related_result["MERGE_COMMENTS"])
+        ]
+        for related_related_mrn in related_related_mrns:
```
davidmiller: Use recursion ? 
(If nesting can happen one level, is there any reason it can't happen a second level?)

intrahospital_api/update_demographics.py
```diff
@@ -194,6 +195,103 @@ def update_if_changed(instance, update_dict):
...
+
+    active_rows = [
+        i for i in mrn_to_row.values()
+        if i["ACTIVE_INACTIVE"] == models.MergedMRN.ACTIVE
+    ]
+
+    # if there is not exactly 1 active row, we don't know what to merge
+    # so do nothing.
+    if not len(active_rows) == 1:
+        return mrn, []
```
davidmiller, line 340: I'm not sure 'do nothing' is exactly the right thing to do - are we aware how often this occurs?

fredkingham:
So the MergedMRN model looks like: 

class MergedMRN(models.Model):
    patient = models.ForeignKey(Patient, on_delete=models.CASCADE)
    mrn = models.CharField(max_length=256)
    merge_comments = models.TextField(blank=True, null=True, default="")
    upstream_merge_datetime = models.DateTimeField(blank=True, null=True)
    our_merge_datetime = models.DateTimeField(blank=True, null=True)

So in the simple case of e.g. MRN 123 has the merge comment "Merged with MRN 345 Oct 17 2016 11:03AM" where 123 is inactive and 345 is active this works. We get 
```py
MergedMRN(
   patient=Patient(345), 
   mrn=123, 
   merge_comment="Merged with MRN 345 Oct 17 2016 11:03AM", 
   upstream_merge_datetime=17/10/2016
  )
```
complex case:
 123, inactive, "Merged with 234 on Oct 17 2016 11:03AM"
 234, inactive, "Merged with 123 on Oct 17 2016 11:03AM, merged with 345 on Oct 18 2018 11:03AM"
 345, active, "Merged with 234 on Oct 18 2018 11:03AM"
 
 We get
 ```py
[
    MergedMRN(
        patient=Patient(345), 
        mrn=123,
        merge_comment="Merged with 234 on Oct 17 2016 11:03AM", 
        upstream_merge_datetime=17/10/2016
   ),
    MergedMRN(
        patient=Patient(345), 
        mrn=234,
        merge_comment="Merged with 234 on Oct 17 2016 11:03AM, merged with 345 on Oct 18 2018 11:03AM" 
        upstream_merge_datetime=17/10/2016
    ),
    MergedMRN(
        patient=Patient(345), 
        mrn=234,
        merge_comment="Merged with 234 on Oct 17 2016 11:03AM, merged with 345 on Oct 18 2018 11:03AM", 
        upstream_merge_datetime=18/10/2016
    ),
  ]
```

The second and third rows are mapping the relationships from mrn 234 to merge comment to 123 and 345 respectively but are a bit useless as its not clear the MRN they are merging. The issue is the upstream datetime is a function of the deconstructed merge comment, while the rest is that mapping of an MRN to an active MRN

fredkingham:
I have decided to change the MergedMRN to be explicitly defined as a row in the Cerner masterfile for an MRN that is inactive. 

This means that the upstream_merge_datetime has a fuzzy meaning as there can be 2 merges for a single MRN into inactive rows. The upstream_merge_datetime is only used for ordering the merged MRNs in the search results so approximate is ok. I've changed the upstream_merge_datetime to be the highest datetime mentioned in the merge comment.

Needs discussion as If you're happy with this approach it would be more sensible to delete upstream_merge_datetime and move that logic into the search results

fredkingham:
complex case:
123, inactive, "Merged with 234 on Oct 17 2016 11:03AM"
234, inactive, "Merged with 123 on Oct 17 2016 11:03AM, merged with 345 on Oct 18 2018 11:03AM, merged with 456 Oct 19 11:03 AM,"
345, inactive, "Merged with 234 on Oct 18 2018 11:03AM, Merged with 456 20 Oct 2018 11:03AM"
456, active, "Merged with 234 Oct 19 11:03 AM, Merged with 345, 20 Oct 2018 11:03AM"

fredkingham:
As discussed, we've removed upstream_merge_datetime

fredkingham:
Todays changes do 2 things.

1. We raise different exceptions on different problems. 
     * MergeException if we should allow the patient to be created but that we should be notified
     * CernerPatientNotFoundException if the patient is not found in the master file. We should not create the patient and raise an error. This can be handled by the calling code.
     * ValueError if there are multiple rows for an MRN in the MasterFile. This should never be the case and to are knowledge has never been the case. We should probably stop the process if this has happened as it means something really wrong has happened to the upstream data.
     
2. `get_or_create_patient` now accepts an rfh_patient argument (default=True). 
We discussed RNOH patients and how they would not exist in the CernerMasterFile. I am assuming they also would not have lab/imaging data? If we pass rfh_patient=False then we allow a patient to be created without checking demographics to make sure they exist and without the loaders running.

This leads us to the situation where create_rfh_patient_from_hospital_number takes an argument called `rfh_patient` patient which is obviously not great but we are planning to rename this function in a future PR where we will add __ and remove the reference to rfh in the name.

PR #1367, opened 2022-12-09
Adds the merge_patient function

fredkingham:
Adds in intrahospital_api.merge_patient.merge_patient that takes two patients and merges one onto the other.

### DIFF
--- a/intrahospital_api/merge_patient.py
+++ b/intrahospital_api/merge_patient.py
@@ -78,7 +78,8 @@ def copy_tagging(old_episode, new_episode):
         ).first()
         if new_tag:
             if new_tag.archived and not old_tag.archived:
-                # if old_tag is not archived consider it the primary
+                # if tag on the old episode is not archived consider
+                # then unarchive it then unarchive it on the new one
                 new_tag.archived = False
                 new_tag.user = old_tag.user
                 new_tag.save()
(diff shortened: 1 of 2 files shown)


intrahospital_api/merge_patient.py
```diff
@@ -0,0 +1,146 @@
...
+    else:
+        qs = subrecord_cls.objects.filter(patient=old_parent)
+    if subrecord_cls._is_singleton:
+        update_singleton(qs[0], new_parent, old_mrn, new_mrn, is_episode_subrecord)
+    else:
+        copy_non_singletons(qs, new_parent, old_mrn, is_episode_subrecord)
+
+
+@transaction.atomic
+def merge_patient(old_patient, new_patient):
```
davidmiller: Suspect t's clearer in the calling code if these are named kwargs

intrahospital_api/merge_patient.py
```diff
@@ -0,0 +1,146 @@
...
+    Copy over any episode categories that do not exist, iterate
+    over subrecord attached to these and add the PreviousMRN
+
+    Elcid native singleton entries to pick the latest but create a reversion history entry for the non-oldest, with a reference to the original_mrn
+    """
+    old_mrn = old_patient.demographics().hospital_number
+    new_mrn = new_patient.demographics().hospital_number
+
+    for patient_subrecord in subrecords.patient_subrecords():
+        if not issubclass(patient_subrecord, models.PreviousMRN):
```
davidmiller: What does this exclude ? Do we definitely catch everything ? OTTOMH is ICUhandover in `load_patient` ?

PR #1368, opened 2022-12-09
Makes get_patients_from_mrns more efficient.

fredkingham:
Only look for MRNs in the MergedMRN table if they have not been found in demographics.


davidmiller:
Is this branch right? Not sure where the conflicts are coming from - perhaps force pushing ?

fredkingham:
Yep its because I force pushed a fix to a typo in a comment 😞

PR #1370, opened 2022-12-12
Changes the discharge summary meds creation to use bulk create

fredkingham:
The average patient has 4 discharge summary medications so using a bulk create will give us a quick and easy speed improvement


davidmiller:
If we want this, it' def. not this branch...

PR #1372, opened 2022-12-12
Merge bed status batch update

fredkingham:
Handle batch bed status creation with merged MRNs

### DIFF
--- a/plugins/admissions/loader.py
+++ b/plugins/admissions/loader.py
@@ -12,7 +12,7 @@ from opal.models import Patient
 
 from elcid.episode_categories import InfectionService
 from elcid.models import Demographics
-from elcid.utils import get_patients_from_mrns
+from elcid.utils import find_patients_from_mrns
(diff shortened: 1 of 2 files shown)


plugins/admissions/loader.py
```diff
@@ -354,29 +355,28 @@ def load_bed_status():
...
+            )
+
     with transaction.atomic():
 
         BedStatus.objects.all().delete()
 
         for bed_data in status:
-            bed_status = BedStatus()
+            patient = mrn_to_patient.get(bed_data["Local_Patient_Identifier"])
+            if not patient:
```
davidmiller, line 377: I think what `if not patient` really means is `if bed_data["Local_Patient_Identifier"] == None || bed_data["Local_Patient_Identifier"] == '' || bed_data["Local_Patient_Identifier"].lstrip('0') == '' ` but done implicitly by another function in another file that throws away data it doesn't like? 

I predict that won't be immediately obvious in 6months - as a minimum we should explain how it happens

PR #1373, opened 2022-12-12
Merge batch ipc status

fredkingham:
Makes the batch IPC status load work with merged MRNs


plugins/ipc/management/commands/load_ipc_status.py
```diff
@@ -84,17 +85,23 @@ def handle(self, *args, **options):
...
+
         for row in upstream_result:
-            if row['Patient_Number'] in inpatients:
-                patient = Patient.objects.get(demographics__hospital_number=row['Patient_Number'])
+            patient = mrn_to_ipc_patients.get(row['Patient_Number'])
+
+            if not patient:
+                continue
 
+            if patient.id in inpatient_ids:
```
davidmiller: We might have done this for some testing purpose, but this (and L88-90) above shouldn't have crept into a release branch - suggest we take it out now in case we forget it's here

fredkingham: Are we talking about either

a) Removing IPC creation
b) Changing it so that regardless of bed status IPC statuses are created.

fredkingham:
This loader is only going to be run once. 

Therefore we want to load for everyone and create episodes without constraining it to bed status.

Remove the filtering by bed status.

plugins/ipc/loader.py
```diff
@@ -26,12 +26,22 @@ def load_all_extra_ipc_patients():
             Q_GET_DISTINCT_MRNS_FOR_TEST,
             params={'test_name': test.TEST_NAME}
         )
-        all_tested_mrns.update([r['Patient_Number'] for r in tested_mrns])
-
-    all_mrns = set(
+        # elcid does not use zero prefixed MRNs but the lab test
```
davidmiller: This should probably be signalled harder, but this whole function is more of a development/testing thing, related to the initial development of `ipc.models.InfectionAlert` - that's paused for now, I think we'd just 'do this differently'. 
We should probably delete the whole function and anywhere it's called rather than add complexity

plugins/ipc/management/commands/load_ipc_status.py
davidmiller: As far as I can tell, the behaviour here is still not as we want it - am I missing where this script creates the patient?

fredkingham:
This needs further conversation see https://github.com/orgs/openhealthcare/projects/1/views/1?pane=issue&itemId=22173238

plugins/ipc/loader.py
fredkingham: cool, in that case we aren't using anything in plugins.ipc.loader so I have deleted it.

fredkingham:
This still needs to be changed to create patients

PR #1374, opened 2022-12-12
Merge icu handover

fredkingham:
Makes the batch ICU Handover load allow for merged patients.

### DIFF
--- a/plugins/icu/loader.py
+++ b/plugins/icu/loader.py
@@ -9,6 +9,7 @@ from opal.models import Patient
 
 from elcid.models import Demographics
 from elcid.episode_categories import InfectionService
+from elcid import utils
 from intrahospital_api.apis.prod_api import ProdApi as ProdAPI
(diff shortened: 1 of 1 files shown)


plugins/icu/loader.py
```diff
@@ -59,67 +60,73 @@ def load_icu_handover():
...
+    upstream_mrns = [result['Patient_MRN'] for result in results]
+    mrn_to_patients = utils.find_patients_from_mrns(upstream_mrns)
+
+    for upstream_mrn in upstream_mrns:
+        if upstream_mrn not in mrn_to_patients:
+            # ICU handover MRNs can have preceding zeros, elCID does not use zero
+            # prefixes as we match the upstream masterfile table
+            mrn = upstream_mrn.lstrip('0')
+            if mrn:
+                patient = create_rfh_patient_from_hospital_number(
```
davidmiller, line 72: You don't think this should be async=True?

PR #1377, opened 2022-12-13
Changes the transfer history model to be inline with upstream table changes

fredkingham:
Upstream have added in fields and removed fields. This changes the
transfer history loader to save those fields


plugins/admissions/models.py
```diff
@@ -270,8 +273,11 @@ class Meta:
         'ROOM'                                  : 'room',
         'BED'                                   : 'bed',
         'TRANS_HIST_TRANSFER_REASON'            : 'transfer_reason',
-        'CREATED_DATE'                          : 'created_datetime',
         'UPDATED_DATE'                          : 'updated_datetime',
+        'TRANS_UPDATED'                         : 'trans_updated_datetime',
```
davidmiller: These aren't the fieldnames it'll set them as properties and never save them - imagine we didn't test on test and check data ?

PR #1378, opened 2022-12-13
Adds the primary diagnosis to the default tb epr message

fredkingham:
Adds in the primary diagnosis to the default TB EPR message.

### DIFF
--- a/plugins/tb/tests/test_epr.py
+++ b/plugins/tb/tests/test_epr.py
@@ -1,0 +1,50 @@
+from django.contrib.auth.models import User
+from opal.core.test import OpalTestCase
+from elcid.models import Diagnosis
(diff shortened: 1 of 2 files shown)


plugins/tb/tests/test_epr.py
```diff
@@ -0,0 +1,50 @@
...
+			last_name="Watson",
+			email="s.watson@nhs.net",
+			username="s.watson"
+		)
+		self.pc = self.episode.patientconsultation_set.create(
+			created_by=self.doctor
+		)
+
+
+	def test_with_primary_diagnosis(self):
```
davidmiller: These test names are the opposite way round to the test contents I think?

PR #1379, opened 2022-12-13
Removes the epr padding of TB treatments from doctors advice

fredkingham:
At present Stop dates are appearing on the next line so remove the padding so this no longer happens.


fredkingham:
Closing this in favour of https://github.com/openhealthcare/elcid-rfh/pull/1383

PR #1382, opened 2022-12-14
Adds the check_transfer_histories management command

fredkingham:
This looks at counts between elCID and upstream as a sanity check that the transfer history loader is working.

I think this is a useful thing to have in our code base but I understand if you think its just noise.

There is no cron job for it, I was calling it after the fetch_transfer_histories ran.


fredkingham:
Closing this. Its great but my assumption is that for this release we will not be creating all patients with transfer histories so the test is not valid at this time.

PR #1387, opened 2022-12-21
Adds frequency to the treatment display.

fredkingham:
This is used by TB Medications and Other Medications on the TB page.]

We force a line break before it because otherwise it will go onto the next line breaking in the middle.

Screenshot 2022-12-21 at 10 14 20


fredkingham:
change it to 
{start_date } - {end_date} {e.g. day 6/30}
{drug_name} {frequence} {route}

fredkingham:
As discussed, this is reorderd so if there is an end date, the regime data is all on one line.

Screenshot 2023-01-09 at 15 33 12

PR #1397, opened 2023-01-20
Removes useless comment tags

fredkingham:
These comment tags don't work so we may as well remove them.

Django supports 
```{# one line comment #}```

or 
```
{% comment %}
this is commented
{% endcomment %}
```
{# comment #} something {# endcomment #} is not a thing and probably just confuses things 

https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#templatetag


davidmiller:
Is there a related ticket to remove that permission ?

fredkingham:
There can be, I'm not sure that blocks this.

fredkingham:
Remove the if clause in the template instead

davidmiller:
Is this the right branch?

PR #1399, opened 2023-02-01
Updated contact tracing flow for location history

davidmiller:
Preview of inpatient contact tracing flow (Location history, contacts, bed history, slice detail)

The instances of scrollable-table-container likely need replacing with some position fixed +fixed height options at varying heights

Requires a fully sync'd IPC status, location history, bed_status to preview


fredkingham:
A few suggestions so far

Long locations do weird things to the buttons http://192.168.21.203/#/admissions/encounter/129079734/contacts/

You also maybe want to leave transfer reason as blank rather than "None"

This page is not working http://192.168.21.203/#/admissions/encounter/116200754/contacts/ (could be an issue with the data sync?)

PR #1406, opened 2023-02-08
Changes us to use time.montonic in the timing utils

fredkingham:
time.time uses the system clock which has various issues, sometimes it can go down for example with leap seconds.

We do not use time.monotonic_ns as it does not exist on all platforms.

### DIFF
--- a/elcid/utils.py
+++ b/elcid/utils.py
@@ -7,7 +7,7 @@ import logging
 import os
 import re
 import sys
-from time import time
+import time
 
 from django.utils import timezone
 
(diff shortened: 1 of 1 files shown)


davidmiller:
Has this caused actual issues? Do we definitely care?

fredkingham:
It doesn't cause issues its mostly used in logging but we do use it (and I find it very useful).

We might as well do it right.

fredkingham:
Check to see if we want to use perf counter. Also this is low priority

PR #1422, opened 2023-03-07
Special cases icp status in the merge patient

fredkingham:
For the merge patient we cannot rely on created/updated timestamps instead we should use the most recent infection date.

This involves a unit test that should error if additional infection dates are added or if infection dates are removed.

### DIFF
--- a/intrahospital_api/merge_patient.py
+++ b/intrahospital_api/merge_patient.py
@@ -41,7 +41,6 @@ PATIENT_RELATED_MODELS = [
     icu_models.ICUHandoverLocationHistory,
     intrahospital_api_models.ExternalDemographics,
     intrahospital_api_models.InitialPatientLoad,
-    ipc_models.IPCStatus,
     opal_models.InpatientAdmission,
     opal_models.PatientRecordAccess,
     rnoh_models.RNOHDemographics,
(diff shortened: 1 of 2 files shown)


intrahospital_api/merge_patient.py
```diff
@@ -248,9 +270,45 @@ def move_non_singletons(subrecord_cls, old_parent, new_parent, old_mrn):
...
+        getattr(old_ipc_status, field) for field in IPC_DATE_FIELDS if getattr(old_ipc_status, field) is not None
+    ]
+
+    new_ipc_status = new_patient.ipcstatus_set.get()
+    new_ipc_dates = [
+        getattr(new_ipc_status, field) for field in IPC_DATE_FIELDS  if getattr(new_ipc_status, field) is not None
+    ]
+
+    # If there are no dates on either the old or new statuses do nothing.
+    if len(old_ipc_dates) == 0 and len(new_ipc_dates) == 0:
```
davidmiller, line 290: I'm not convinced this behaviour is definitely what we want,
e.g. what if there are comments & / or boolean flags but no dates on one but not the other? 

Also: Did we look at using `Last_Updated` instead of all these date fields ?

fredkingham:
#### Re: last updated.

Last updated seems to mirror date inserted.

The data for this is on the test server in ipc_rows.csv. Looking at ids 16321 and 18047 (rows 10 and 11), I think we want to use line 11.

I think we want line 11 because it has the most recent covid positive, however it has an earlier date inserted and last updated.

fredkingham:
#### Re: what happens if date is null but the boolean field is not.

I don't _think_ this ever seems to be the case.

```

def not_date():
    query = "SELECT TOP(1) * FROM ElCid_Infection_Prevention_Control_View WHERE {non_date_field} is not null and {non_date_field} <> '' AND {date_field} is null"
    api = ProdAPI()
    stuff = (
        ("Acinetobacter", "Acinetobacter_LPS_Date"),
        ("CJD", "CJD_LPS_Date"),
        ("C_DIFFICILE", "C_DIFFICILE_LPS_Date"),
        ("Candida_auris", "Candida_auris_date"),
        ("Carb_resistance", "Carb_resistance_LPS_Date"),
        ("Contact_of_Acinetobacter", "Contact_of_Acinetobacter_LPS_Date"),
        ("Contact_of_Candida_auris", "Contact_of_Candida_auris_date"),
        ("Contact_of_Carb_resistance", "Contact_of_Carb_resistance_LPS_Date"),
        ("Contact_of_Covid19", "Contact_of_Covid19_Date"),
        ("Covid19", "Covid19_Date"),
        ("MRSA", "MRSA_LPS_Date"),
        ("MRSA_neg", "MRSA_neg"),
        ("Multi_drug_resistant_organism", "Multi_drug_resistant_organism_date"),
        ("Other_Type", "Other_Date"),
        ("Reactive", "Reactive_LPS_Date"),
        ("VRE", "VRE_LPS_Date"),
        ("VRE_Neg", "VRE_Neg_LPS_Date"),
    )
    for row in stuff:
        result = api.execute_hospital_query(query.format(non_date_field=row[0], date_field=row[1]))
        if result:
            import pdb; pdb.set_trace()
        pass
```
never pauses

fredkingham:
closing this as it needs further thought

PR #1423, opened 2023-03-08
168 disallow multiple user email addresses

fredkingham:
Changes the admin user form to raise an error if a user is added with the same email as a preexisting user.
Changes create_covid_user to raise an error if a user is added with the same email as a preexisting user.

Screenshot 2023-03-08 at 13 43 17

### DIFF
--- a/elcid/admin.py
+++ b/elcid/admin.py
@@ -15,8 +15,8 @@ admin.site.unregister(User)
 
 
 class ElcidUserAdmin(opal_admin.UserProfileAdmin):
-	form = forms.ElcidUserAdminChangeForm
-	add_form = forms.ElcidUserAdminCreationForm
+    form = forms.ElcidUserAdminChangeForm
+    add_form = forms.ElcidUserAdminCreationForm
 
 admin.site.register(User, ElcidUserAdmin)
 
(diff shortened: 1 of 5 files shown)


elcid/admin.py
```diff
@@ -1,10 +1,24 @@
...
+from elcid import forms
 
 from elcid.models import BloodCultureIsolate
 
 
+admin.site.unregister(User)
+
+
+class ElcidUserAdmin(opal_admin.UserProfileAdmin):
+	form = forms.ElcidUserAdminChangeForm
```
davidmiller: We're doing tabs now?

elcid/forms.py
```diff
@@ -1,10 +1,44 @@
...
 
 class BulkCreateUsersForm(forms.Form):
     """
     Form for uploading a CSV of users to add.
     """
     users = forms.FileField()
+
+
+class ElcidUserAdminCreationForm(user_forms.UserCreationForm):
+    def clean_email(self):
```
davidmiller: Would you have these here rather than `admin.py` ?

fredkingham: So its elcid/forms.py to mirror the structure of how django does it. I don't mind either way but that was the thinking.

fredkingham: As discussed lets move the forms into admin.py

PR #1425, opened 2023-03-10
Changes transfer history/encounter batch loads

fredkingham:
Changes the encounter and transfer history batch loads to work with mergedMRNs.

As part of this I've remove the encounter patient creation from load_excounters_since into update_encounters_from_query_result as otherwise we are doing the demographics lookup twice.


fredkingham:
This still does not deal with merged patients (or at least it does but it will error)

plugins/admissions/loader.py
```diff
@@ -100,29 +100,26 @@ def cast_to_encounter(encounter, patient):
...
-            patient
-        )
-
+    mrn_to_patients = find_patients_from_mrns(mrns)
     to_create = []
     for row in rows:
-        for patient in mrn_to_patients[row['PID_3_MRN'].strip()]:
-            to_create.append(cast_to_encounter(row, patient))
+        mrn = row['PID_3_MRN']
+        if mrn not in mrn_to_patients:
```
davidmiller, line 114: Is this how these work? doesn't the way `find_patients_from_mrns` work mean that you have to check again for leading 0s, Nones, empty strings etc?

plugins/admissions/loader.py
```diff
@@ -100,29 +100,26 @@ def cast_to_encounter(encounter, patient):
...
-    ).prefetch_related('demographics_set')
-
-    mrn_to_patients = defaultdict(list)
-    for patient in patients:
-        mrn_to_patients[patient.demographics_set.all()[0].hospital_number].append(
-            patient
-        )
-
+    mrn_to_patients = find_patients_from_mrns(mrns)
     to_create = []
```
davidmiller: I would have this variable as `encounters_to_create` given the creation of patients nearby

plugins/admissions/loader.py
```diff
@@ -257,25 +242,23 @@ def create_patients(mrns):
...
-    ]).select_related('patient')
-    for demo in demos:
-        mrn_to_patients[demo.hospital_number].append(demo.patient)
+    from intrahospital_api.loader import create_rfh_patient_from_hospital_number
+    mrns = [i['LOCAL_PATIENT_IDENTIFIER'] for i in some_rows]
+    mrn_to_patients = find_patients_from_mrns(mrns)
+
+    for mrn in mrns:
+        if mrn not in mrn_to_patients:
+            mrn_to_patients[mrn] = create_rfh_patient_from_hospital_number(
```
davidmiller: Similar.. what if it's e.g. None?

plugins/admissions/loader.py
```diff
@@ -257,25 +243,35 @@ def create_patients(mrns):
...
+    from intrahospital_api.loader import create_rfh_patient_from_hospital_number
+    # Remove rows where the MRN is None, an empty string or only 000s
+    some_rows = [i for i in unfiltered_rows if is_valid_mrn(i['LOCAL_PATIENT_IDENTIFIER'])]
+    mrns = [i['LOCAL_PATIENT_IDENTIFIER'] for i in some_rows]
+    mrn_to_patients = find_patients_from_mrns(mrns)
+
+    for mrn in mrns:
+        if mrn not in mrn_to_patients:
+            mrn_to_patients[mrn] = create_rfh_patient_from_hospital_number(
+                mrn, InfectionService, run_async=False
```
davidmiller: I'm not entirely sure why we're telling this to run_async=False here?

plugins/admissions/loader.py
```diff
@@ -100,29 +100,27 @@ def cast_to_encounter(encounter, patient):
 
 
 def update_encounters_from_query_result(rows):
+    rows = [i for i in rows if is_valid_mrn(i['PID_3_MRN'])]
+    from intrahospital_api.loader import create_rfh_patient_from_hospital_number
     upstream_ids = [i['ID'] for i in rows]
     existing_encounters = Encounter.objects.filter(upstream_id__in=upstream_ids)
     existing_encounters_count = existing_encounters.count()
     logger.info(f'{existing_encounters_count} existing encounters will be removed')
     mrns = [i['PID_3_MRN'].strip() for i in rows if i['PID_3_MRN'].strip()]
```
davidmiller: What's the reason for this? 
As implemented, find_patients_from does loads of stripping itself, and is_valid_mrn on L103 filters out the need for the if clause?

fredkingham: find_patients_from does do the stripping but the `create_rfh_patient_from_hospital_number` does not so if its a new MRN this would mean that we could still create a patient with MRN 0000

That said, it is superfluous because of line 103.

 It would not be if there were any MRNs in the upstream encounter table with prefixed spaces but there aren't.

davidmiller: My comment was 'no need for strips, no need for python truthiness test' 
We lost one but not the other.
Are you sure the if clause is required? What does it do ?

fredkingham: I think this is resolved so what its doing is clearer, also I think it was sligly confused before. 

We only need to do checks and strips befor creating patients.

PR #1428, opened 2023-03-13
Adds a check management command to make sure our merging is working

fredkingham:
Adds a managment command that makes sure...

* We do not have MRNs that are both in Demographics.hospital_number and MergedMRN.

* We do not have MRNs that are marked as active and merged upstream that do not have any merged MRNs

* We do not have MRNs that are marked as inactive upstream and active in our system.


intrahospital_api/management/commands/check_merged_patients.py
```diff
@@ -0,0 +1,88 @@
...
+
+def get_all_merged_patients():
+    api = ProdAPI()
+    return api.execute_hospital_query(GET_ALL_MERGED_PATIENTS)
+
+
+def check():
+    our_merged_mrns = set(elcid_models.MergedMRN.objects.values_list('mrn', flat=True))
+    our_demographics_mrns = set(elcid_models.Demographics.objects.all().values_list('hospital_number', flat=True))
+    our_merged_demographics_mrns = set(elcid_models.Demographics.objects.all().values_list(
```
davidmiller: Is this definitely different from the variable above? 
(I might be code blind)

davidmiller:
What's the context this is expected to execute & how do we know when it's found something?
Surely it should either raise exceptions _or_ send to logger.error not both?

If this is useful it's as something running automatically after the latest merge, so with a try, catch, email block...

In that context I'd add it to a cron job with an && (0 exit code) and remove 'thresholds'... 
(Unless there's a reason the threshold cases will always exist?)

Also, not sure the checks in the code match the list in the description here or the docstring? (can they all match)

fredkingham:
These are all good points.

The reason for the threshold is that we know there are flawed upstream rows so we need to allow for these. I've changed it to explicitly ignore them based on id.

How this runs, I had envisioned it running once a week, just as a check sum to make sure everything works as expected but I think we should talk about this.

fredkingham:
This mgmt command takes ~4 seconds to run

intrahospital_api/management/commands/check_merged_patients.py
```diff
@@ -0,0 +1,232 @@
+"""
+Checks that our Merged MRN cron job is running.
```
davidmiller: That's not strictly what it does? 
that would look like `ps aux | grep merge_patients_since` ?

intrahospital_api/management/commands/check_merged_patients.py
```diff
@@ -0,0 +1,232 @@
...
+    our_demographics_mrns = set(elcid_models.Demographics.objects.all().values_list('hospital_number', flat=True))
+
+    # ALL active MRNs which have an inactive MRN
+    our_merged_demographics_mrns = set(elcid_models.Demographics.objects.filter(
+        patient_id__in=set(elcid_models.MergedMRN.objects.values_list('patient_id', flat=True).distinct())
+    ).values_list(
+        'hospital_number', flat=True
+    ))
+    intersection = our_merged_mrns.intersection(our_demographics_mrns)
+    if len(intersection) > 0:
```
davidmiller, line 186: Is this check listed in any of the docstrings? I can't see it

plugins/monitoring/management/commands/check_merged_patients.py
```diff
@@ -0,0 +1,266 @@
...
+]
+
+
+GET_ALL_MERGED_PATIENTS = """
+    SELECT ID, PATIENT_NUMBER, ACTIVE_INACTIVE, MERGE_COMMENTS FROM CRS_Patient_Masterfile
+    WHERE MERGED = 'Y'
+"""
+
+
+def send_email(title):
```
davidmiller: don't you want this to be utils now?

PR #1429, opened 2023-03-13
Merge all patients 1

fredkingham:
Adds the managment commands `merge_all_patients` and `merge_patients_since` to merge all patients and just the patients in the last couple of days respectively.

This is done by the new `update_demographics.check_and_handle_upstream_merges_for_mrns` which takes in a bunch of MRNs and checks which and what need merging and then does the necessary.


fredkingham:
Moved back into in progress as it requires https://github.com/openhealthcare/elcid-rfh/pull/1440 to be merged.

fredkingham:
Needs a cron job

fredkingham:
Adds in the merge_dup_episodes command. We have a number of patients who have duplicate episode categories, the merge_patient command expects there to be only one episode of each category, this is so that it can know where to copy episode subrecords.

fredkingham:
Added the merge patients since job to the sync_demographics cron file. I've started adding cron jobs to the existing files rather than using new cron files each time. Roughly these groupings are based around packages, e.g. admissions has fetch_bed_status in it. There is no elcid cron job so I've put it in sync demographics

intrahospital_api/management/commands/merge_all_patients.py
```diff
@@ -0,0 +1,166 @@
...
+                old_mrn,
+            )
+    old_patient.delete()
+    new_patient.mergedmrn_set.filter(mrn=old_mrn).update(
+        our_merge_datetime=timezone.now()
+    )
+    merge_patient.updates_statuses(new_patient)
+
+
+@transaction.atomic
```
davidmiller, line 79: I either don't think nested transactions work the way they're written in this file, or don't understand the intention...

intrahospital_api/management/commands/merge_all_patients.py
```diff
@@ -0,0 +1,166 @@
...
+def check_and_handle_upstream_merges_for_mrns(mrns):
+    """
+    Takes in a list of MRNs.
+
+    Filters those not related to elCID.
+
+    If they are inactive, creates a Patient for the
```
davidmiller: Is this accurate to the implementation isn't it filtered on active in the query?

intrahospital_api/management/commands/merge_all_patients.py
fredkingham: I don't _think_ it matters but we don't need 2 transaction atomics.

fredkingham:
Some stats..
#### Looking at all merge timestamps in everything merged upstream.
Since 2015 there have been 39 times where there have been more than 100 merges 
All of these were less than 300 merges except on 2/3/2022 where there were 9081 merges

#### Looking at fact timestamps 
Its run 152 times on test. 
104 of these times took less than 1 mins.
9 times over ten mins
3 times at 25 mins
1 time at 34 mins 
1 time at 40 mins.

Not sure what was happening in those situations, there's no massive jump in merges when there's a jump in time taken

fredkingham:
This is going to be changed to send an email if we get an abnormally large amount back from the server.
It will also be changed to use https://github.com/openhealthcare/elcid-rfh/pull/1470 when merged.
Finally change the message sent out to reflect its the large amount of MergedMRNs we are worried about.

fredkingham:
The hours when there aremerges ...

8 3211
9 9084
10 7499
11 7999
12 6802
13 3878
14 15561
15 6298
16 5154
17 449
18 167

ie we don't need to run it over night.

fredkingham:
~~The 200,000 threshold came from the fact that that is over more than we would expect to return in a month (last month we returned 169,376).~~ <-- was the count for all (i.e. including non merged) MRNs. 

The total number of merged ='Y' patients updated in the last month is 4582

I test how it would cope, changing the command to run for the last month and it finished in under a minute

intrahospital_api/management/commands/merge_patients_since.py
```diff
@@ -0,0 +1,38 @@
...
+import time
+import traceback
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+from opal.models import Patient
+from intrahospital_api import update_demographics, logger, constants
+from plugins.monitoring.models import Fact
+from elcid import utils
+
+UNEXPECTED_VOLUMES_THRESHOLD = 200000
```
davidmiller: I was thinking more like 500?
Given your test, maybe start with 1000, and put it in an actual setting so we don't need to redeploy to change it ?

intrahospital_api/management/commands/merge_patients_since.py
```diff
@@ -0,0 +1,38 @@
...
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        try:
+            start = time.time()
+            since = timezone.now() - datetime.timedelta(2)
+            upstream_merged_mrns = update_demographics.get_all_merged_mrns_since(since)
+            if len(upstream_merged_mrns) > UNEXPECTED_VOLUMES_THRESHOLD:
+                utils.send_email(
+                    f"{len(upstream_merged_mrns)} being processed by merge_patients_since",
```
davidmiller: You're missing the word MRN because you've read it in the variable name?

intrahospital_api/update_demographics.py
```diff
@@ -220,13 +237,161 @@ def get_mrn_and_date_from_merge_comment(merge_comment):
...
+    active_mrn_and_merged_dicts = []
+    for active_mrn, merged_dicts in unique_active_mrn_and_merged_dicts:
+        if mrn_in_elcid(active_mrn):
+            active_mrn_and_merged_dicts.append((active_mrn, merged_dicts,))
+        elif len([i for i in merged_dicts if mrn_in_elcid(i["mrn"])]) > 0:
+            active_mrn_and_merged_dicts.append((active_mrn, merged_dicts,))
+    logger.info('Generating merged MRNs')
+    to_create = []
+    for active_mrn, merged_dicts in active_mrn_and_merged_dicts:
+        active_patient = Patient.objects.filter(
```
davidmiller: May be missing something, but I think this spelling has you looping more times and setting up more datastructures than the way I would have spelled it...

```python
active_mrns_seen = set()
for mrn in mrns:
    active, merged_data = get_active_mrn_and_merged_mrn_data(mrn, mrn_to_upstream_merge_data)
    if active in active_mrns_seen:
        # The list passed into this function potentially contains all nodes in a merge graph
        continue 
    if not mrn_in_elcid(active):
        if not any(mrn_in_elcid(i['mrn']) for i in merged_data):
            continue # no node in this graph is part of the elCID cohort
    active_patient = Patient.objects.filter(demographics__hospital_unmber=active).first()
```

intrahospital_api/update_demographics.py
```diff
@@ -220,13 +237,161 @@ def get_mrn_and_date_from_merge_comment(merge_comment):
...
+            mrn__in=merged_mrns
+        )
+        unmerged_patients = Patient.objects.filter(
+            demographics__hospital_number__in=merged_mrns
+        )
+        patients_now_merged = set()
+
+        # If we have patients that are inactive we need to do a merge.
+        if len(unmerged_patients) > 0:
+            if not active_patient:
```
davidmiller: Do you need to do this here as well as L317?

intrahospital_api/update_demographics.py
```diff
@@ -220,13 +237,161 @@ def get_mrn_and_date_from_merge_comment(merge_comment):
...
+                active_patient, _ = loader.get_or_create_patient(
+                    active_mrn,
+                    elcid_episode_categories.InfectionService,
+                    run_async=False
+                )
+            for unmerged_patient in unmerged_patients:
+                if active_patient:
```
davidmiller: Not sure I know what this if statement is for?

intrahospital_api/management/commands/merge_patients_since.py
fredkingham: Sounds good, over the last month there have been 3 days were last_updated > 300, 1 instance where its been above 400 so I think 1000 is better than 500

intrahospital_api/update_demographics.py
fredkingham: Sure, I thought multiple for loops made it more readable, because each did one thing. I'm happy with either.

fredkingham:
So there is an issue with this...

When we call merge_patient.merge_patient we have not created MergedMRNs.

merge_patient.merge_patient implicitly assumes these MergedMRN for the patient exists when it calls load_patient. 

Without MergedMRNs the upstream data will not load in the data for the MergedRNs.

PR #1432, opened 2023-03-15
Adds a "Failing Tests" label when a PR is failing tests

fredkingham:
Adds a label to a PR that fails tests so that it shows in the projects screen.

Screenshot 2023-03-15 at 17 36 58


fredkingham:
So the agreed call is this would be fine if we could also automatgically remove it when tests pass

PR #1433, opened 2023-03-16
Ipc logging

fredkingham:
IPC at present does no logging. This adds the logging.

load_ipc_status does not send any emails if it fails. This changes the mgmt command to email if it does fail.

**Potentially contentious** At the moment there is an MRN in this table that is not present in the cerner table. I have changed the command to email in this case and carry on. Is this what we want? The alternative is that we create a non RFH patient but my assumption is that this is an error in the ipc status and therefore we should just skip them.

### DIFF
--- a/elcid/settings.py
+++ b/elcid/settings.py
@@ -331,6 +331,11 @@ LOGGING = {
             'level': 'INFO',
             'propagate': True,
         },
+        'ipc': {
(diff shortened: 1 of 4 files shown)


fredkingham:
1. Can we look at patients that are in the IPC table that are not in the cerner table
2. Can we create a seperate PR that brings in logging into IPC

PR #1436, opened 2023-03-16
Explicit error messages

fredkingham:
Includes the MRN in the error log statement so that we can then more easily find out what is causing the issue.

Adds an additional unit test.


fredkingham:
Added needs descussion as I'm not sure why this was in in progress, it could well be that I had forgotten to move it into code review

PR #1439, opened 2023-03-23
Adds the non zero prefixed MRN database constraint

fredkingham:
Adds a database constraint to make it so that it is impossible to change an MRN so that it has a zero prefix.

Tested with csv imports and bulk updates.


davidmiller:
Does the constraint work if we deploy it to a database that hasn't yet been cleansed of 0prefixes? 
(If not, does this need to point at a different branch... )

elcid/models.py
```diff
@@ -75,6 +75,12 @@ def age(self):
 
     class Meta:
         verbose_name_plural = "Demographics"
+        constraints = [
+            models.CheckConstraint(
+                check = ~models.Q(hospital_number__startswith=0),
```
davidmiller, line 82: I think I landed at either a different spelling or a comment explaining how _bitwise_ not works here...

PR #1442, opened 2023-04-07
Makes find_patients_from_mrns handle Nones

fredkingham:
The bed status table has None MRNs, handle this without errorring


elcid/utils.py
```diff
@@ -96,7 +96,7 @@ def find_patients_from_mrns(mrns):
     e.g. 000 will be removed.
     """
     cleaned_mrn_to_mrn = {
-        i.strip().lstrip('0'): i for i in mrns if i and i.strip().lstrip('0')
+        i.strip().lstrip('0'): i for i in mrns if i is not None and i.strip().lstrip('0')
```
davidmiller, line 99: Are you sure this change is required?
(Besides it being close to the upper bound for happy dictionary comprehensions, why did the test on L63 pass before this change?)

fredkingham:
Your right, this should be covered by *if i* and is already covered by unit tests

PR #1443, opened 2023-04-07
Use get_or_create_patient in the IPC status table

fredkingham:
The IPC Status tables has multiple rows per MRN, make sure we don't error when loading these in.

### DIFF
--- a/plugins/ipc/management/commands/load_ipc_status.py
+++ b/plugins/ipc/management/commands/load_ipc_status.py
@@ -100,7 +100,10 @@ class Command(BaseCommand):
                 if len(mrn) == 0:
                     continue
                 else:
-                    patient = loader.create_rfh_patient_from_hospital_number(
+                    # The table contains multiple results for the same
+                    # MRN so get or create in case they have already
+                    # recently been created.
+                    patient, _ = loader.get_or_create_patient(
                         mrn, elcid_episode_categories.InfectionService
                     )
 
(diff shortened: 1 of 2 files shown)


plugins/ipc/management/commands/load_ipc_status.py
```diff
@@ -100,7 +100,10 @@ def handle(self, *args, **options):
                 if len(mrn) == 0:
                     continue
                 else:
-                    patient = loader.create_rfh_patient_from_hospital_number(
+                    # The table contains multiple results for the same
+                    # MRN so get or create in case they have already
+                    # recently been created.
+                    patient, _ = loader.get_or_create_patient(
```
davidmiller, line 106: Wouldn't this mean that we care about the ordering of the results in some way? 
this overwrites with the arbitrarily last - is that the behaviour we want?

fredkingham: So there is an ongoing discussion on how to merge ipc status https://github.com/openhealthcare/Development/issues/166

you're right there is an arbitrary behaviour that already exists, this is more so that the we can run the ipc command as with the new create_rfh_patient_from_hospital_number error handling this will blow up and so the command can't be run.

2 options the way I see it...
1) We leave it breaking as a reminder its somethings we have to fix
2) We take add the merge issue to 166 and move on. 

I'm happy with either

fredkingham:
I'm removing the Cerner Merge label as this is not required for that release.

fredkingham:
Marking as needs discussion as https://github.com/openhealthcare/Development/issues/166 is in needs discussion

PR #1446, opened 2023-04-11
Tb birth place

fredkingham:
Moves Demographics.birth_place to Nationality.birth_place.

I haven't changed any of the advanced search as there is already Demographics.gp_practice code that is there and not populated.


plugins/tb/migrations/0060_auto_copy_over_birth_place.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tb', '0059_auto_20230411_1627'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            migrate_forwards
```
davidmiller: Will this now stack trace if I try to go backwards?

fredkingham: It did, I've actually changed it so that the reverse will copy data back from the tb nationality model as presumably that's what we want to theoretically undo this change (although we would probs do it in a seperate migration if so)

PR #1448, opened 2023-04-12
Makes create_rfh_patient_from_hospital_number not error if the patient is not found upstream

fredkingham:
If a patient has an MRN in the lab systems we should not necessarily error if they are not found in the upstream system.

The question is how you want this handled in regards to IPC. Do you think its appropriate in all cases?


fredkingham:
As discussed we don't want to do this, we feel we should handle this on a loader by loader basis

PR #1450, opened 2023-04-13
Fixes a bug where get_or_create_patient with rfh_patient = False errored

fredkingham:
intrahospital_api.loader.get_or_create_patient should not error when rfh_patient is False but should handle it.

Also adds a couple of unit tests to make sure the rfh and non rfh cases are covered.


fredkingham:
closing this, as we just want to remove this flag from the time being. https://github.com/openhealthcare/elcid-rfh/pull/1451

PR #1454, opened 2023-04-18
New patient demographics search with merged mrn

fredkingham:
* Renames load_demographics to search_upstream_demographics as this is what it actually is
* Changes search_upstream_demographics so that it checks to see if the mrn has been merged and then returns results if it has
* Changes the add patient search view to check to see if we have a local merged mrn before returning a result
* Changes the add patient step to use get_or_create_rfh_patient if they exist upstream
* Changes the AddAntifungalPatients to create MergedMRNs

### DIFF
--- a/elcid/api.py
+++ b/elcid/api.py
@@ -584,7 +584,7 @@ class DemographicsSearch(LoginRequiredViewset):
             return json_response(dict(status=self.PATIENT_NOT_FOUND))
         else:
             if settings.USE_UPSTREAM_DEMOGRAPHICS:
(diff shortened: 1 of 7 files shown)


fredkingham:
We now use the get_or_create_patient function. This will load data from upstream and on the off chance that the patient has been created somewhere else between the pathway search screen and the save screen, it will mean we do not recreate the patient,.

We still have to call the base pathway save method. This is to create patients where they do not, for whatever reason, exist upstream and to save the subrecord data.

PR #1458, opened 2023-04-20
Removes an unused import in elcid.utils

fredkingham:
We are importing Patient but not using it


fredkingham:
~~Moved to inprogress, tests failing because github actions if 500ing when trying to setup node. I'll take a look tomorrow~~

Now resolved

PR #1459, opened 2023-04-21
Moves the antifungal logic out of a signal into update_from_dict

fredkingham:
We do not want this logic to be called by merge_patient

This will conflict with https://github.com/openhealthcare/elcid-rfh/pull/1456 but probs best kept seperate and the conflict handled


davidmiller:
as discussed, conflicts...

elcid/models.py
```diff
@@ -427,6 +427,12 @@ def update_from_dict(self, data, *args, **kwargs):
             ).first()
             if micro_input:
                 micro_input.delete_self()
+
+        if data.get("reason_for_interaction") == self.ANTIFUNGAL_STEWARDSHIP_ROUND:
+            omodels.Episode.objects.get(id=data['episode_id']).patient.chronicantifungal_set.create(
```
davidmiller: Use self.episode?

PR #1462, opened 2023-04-23
Log before and after merge

fredkingham:
* log the old patient  the new patient before the merge. Log the combined patient after the merge.
* Remove some unused imports in merged_patient
* mock load_patient as otherwise we get errors in the logs when it tries to hit the upstream apis

### DIFF
--- a/intrahospital_api/merge_patient.py
+++ b/intrahospital_api/merge_patient.py
@@ -1,15 +1,12 @@
 from django.utils import timezone
 from intrahospital_api import loader
-from opal.core import subrecords
 from django.db import transaction
 from elcid import models as elcid_models
 from plugins.admissions import models as admission_models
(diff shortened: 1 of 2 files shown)


intrahospital_api/merge_patient.py
```diff
@@ -22,6 +22,9 @@
 
 IGNORED_FIELDS = {"id", "episode", "patient", "previous_mrn"}
 
+# not *.log so it will not be picked up log rotate
```
davidmiller, line 26: have we double checked our configuration of logrotate?

davidmiller:
Merge criteria:

1. merges.txt _literally_ just the json of the patient about to be deleted and a newline.
2. If you want to log the stuff that's currently being logged to the generic log files, go for it [as long as this doesn't include upstream labs - I don't think it does.]

PR #1463, opened 2023-04-26
Changes create_tb_observations to always run for all time

fredkingham:
At the moment there is a bug if a patient is created who has TB tests prior to 3 days ago, they are not added to the MDT list. We may as well run for all time..

Some timings...
It takes 4 mins as is, 11 mins to run for 8 days and 15 mins to run for all time with the current test server load.

So I think we may as well just run for all time, thoughts?


fredkingham:
This came up again.

When we discussed today we said it should run for the last 8 days. I think we should run for all time as the data displayed in the TB MDT list is data for all time not just the last 8 days and as noted above the time difference is not that great.

PR #1465, opened 2023-04-27
Previous mrn modal fix 1

fredkingham:
Fixes 2 bugs...

1. In some cases the previous MRN was not showing, e.g. if there was no updated timestamp.
2. For MicroInput and other forms that use a custom formItem we weren't showing the created/updated/previous_mrn


fredkingham:
Bumping this up the review queue as https://github.com/openhealthcare/elcid-rfh/pull/1467 will depend on it (after I fix tests)

PR #1470, opened 2023-05-17
Send email util

fredkingham:
Adds a util to send admins emails in a less verbose way.

* prefixes the subject with the opal brand name
* logs before and after


fredkingham:
I've stuck this in cerner merges even though it can go in 108 as there will be an amendment to an existing PR (https://github.com/openhealthcare/elcid-rfh/pull/1429) that depends on it.

PR #1479, opened 2023-05-24
Revert "Fixup: TB model flags"

fredkingham:
Reverts openhealthcare/elcid-rfh#1478


fredkingham:
This should not have one into 108, it should have gone into 109

